PR #7, opened 2017-07-19
Fixes to make clifford run in Python 3

jacopoantonello:
I made some changes to be able to use clifford with Python 3. It appears to be running now although I have not done extensive testing. Would this be of interest?

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -1,6 +1,6 @@
-""" 
+"""
 
-Two classes, Layout and MultiVector, and several helper functions are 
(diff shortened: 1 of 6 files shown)


arsenovic:
thanks a lot, i have been putting this off. 

the test suite for this module is inadequate ,but this works on my scripts so far. the travis builds seem to fail, complaining about something missing. perhaps you know the fix?

https://travis-ci.org/arsenovic/clifford

arsenovic:
nevermind i fixed it.  was missing a future dep .

jacopoantonello:
Thanks!

Issue #9, opened 2017-11-01
Refuses to operate on MultiVectors with identical Layout objects


hugohadfield:
Comparison in MultiVector._checkOther compares ids of layout objects rather than contents of the layout instead. This is a problem when operating with instances of the algebra created with copy.deepcopy() applied to the MultiVector as the contents of the layouts will be the same but the ids of the objects will be different. A quick hack fix is to implement an __eq__ and __ne__ method for the layout class:

    def __eq__(self, other):
      return self.__str__() == other.__str__()
    def __ne__(self, other):
      return not self.__eq__(other)

and change MultiVector._checkOther to use != rather than 'is not'.

Obviously getting the layout as a string and comparing these values is not the best serialisation and comparison scheme. Any detailed comparison of the full contents of the class is likely to affect performance somewhat but I think some kind of comparison of core pieces of information from the layout would be worthwhile instead of the id comparison currently used.

arsenovic:
good point. seems to me that all you need to know is the signature to ensure the GA is the same, so we should be able to do `self.layout.sig == other.layout.sig`,
see any problems with this?

hugohadfield:
Sure, so the overwritten __eq__ and __ne__ operatorsw in Layout should become:
    def __eq__(self, other):
      return np.all(self.sig == other.sig)

    def __ne__(self, other):
      return not self.__eq__(other)

arsenovic:
yeah that makes sense to me.

arsenovic:
do you want to submit a PR? or i can do it,

hugohadfield:
Yeah I was just looking at that! Probably easiest if you do it as I have messed up my local copy of the repo quite a lot...

arsenovic:
yeah  i hear ya, no prob

arsenovic:
let me know if this works for you , 
 https://github.com/arsenovic/clifford/commit/deb9648401f07ca6c5f22745d2055eb456095dca

hugohadfield:
Perfect, thanks :)

PR #18, opened 2017-11-21
Hugo/performance

hugohadfield:
Adds numba as a dependency
Speeds up multiplication code via jitting and sparseness
Speeds up high dimensional algebra generation by jitting

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -201,11 +201,11 @@ def _sign(seq, orig):
     sign = 1
     seq = list(seq)
 
-    for i,s_val in enumerate(seq):
-        if s_val != orig[i]:
-            j = seq.index(orig[i])
-            sign = -sign
-            s_val, seq[j] = seq[j], s_val
+    for i in range(len(seq)):
+            if seq[i] != orig[i]:
+                j = seq.index(orig[i])
+                sign = -sign
+                seq[i], seq[j] = seq[j], seq[i]
     return sign
 
 @numba.njit
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -238,15 +238,6 @@ def modify_idx(idx, grade):
         if not _containsDups(idx):
             done = 1
 
-def get_layout_comp_func(layout_in_sig):
-    '''
-    Returns a partially completed function that compares layouts
-    '''
-    @numba.njit
-    def layout_comp(other_layout_sig):
-        return np.all(layout_in_sig == other_layout_sig)
-    return layout_comp
-
 def get_mult_function(mult_table,n_dims):
     ''' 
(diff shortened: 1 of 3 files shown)


arsenovic:
i measure a speedup of 20s->5s on Cl(8)! awesome. 

so this looks like you have removed  methods of layout used in `__init__`,  jitted them,  which is fine with me. you might add some comments to explain this in the code. 

i use the docs as a test, which is really bad,  but it found the following. 

* bug:  the `comp_func` fails due to array broadcasting. i might have broke this when resolving https://github.com/pygae/clifford/issues/9, with https://github.com/arsenovic/clifford/commit/deb9648401f07ca6c5f22745d2055eb456095dca. you can test with `Cl(2)==Cl(3)` 

* you should also add the new numba dep to `readthedocs-environement.yml` for the docs

hugohadfield:
@arsenovic not sure how I can replicate this comp_func error, Cl(2)==Cl(3) returns False as expected, what behaviour were you seeing?

arsenovic:
i get this . 
```
>>>from clifford import Cl
>>>Cl(2)==Cl(3)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-a887d047d2cf> in <module>()
      1 from clifford import Cl
----> 2 Cl(2)==Cl(3)

/home/alex/code/clifford/clifford/__init__.py in __eq__(self, other)
    416 
    417     def __eq__(self,other):
--> 418         return not self.__ne__(other)
    419 
    420     def __ne__(self,other):

/home/alex/code/clifford/clifford/__init__.py in __ne__(self, other)
    419 
    420     def __ne__(self,other):
--> 421         return self.comp_func(other.sig)
    422 
    423     def _genEvenOdd(self):

ValueError: unable to broadcast argument 1 to output array
File "/home/alex/code/clifford/clifford/__init__.py", line 247
```

moble:
I get the same as Alex, unless I comment out the `njit` decorator — which makes me worry that the code itself is kind of wrong.  The signatures being compared here are numpy arrays `[1,1]` and `[1,1,1]`.  If I just compare those directly, I get a `DeprecationWarning`:

```python
>>> import numpy as np
>>> a = np.array([1,1])
>>> b = np.array([1,1,1])
>>> a != b
DeprecationWarning: elementwise != comparison failed; this will raise an error in the future.
```
I guess the reason is that they're different lengths, and numpy is trying to compare each element.  The usual way to do this comparison would be with `np.array_equal`, which might be faster than the jitted code anyway.  That might be worth a try.

hugohadfield:
@moble I wasn't aware numpy had that function! @arsenovic this should now hopefully be fixed, shall I merge to master?

arsenovic:
works for me

PR #19, opened 2017-12-22
Added a njitted adjoint function

hugohadfield:
Allows the user to write functions such as:
`
def apply_rotor(R,mv):
    return cf.MultiVector(layout,layout.gmt_func(R.value, layout.gmt_func( mv , layout.adjoint_func(R.value))))
`
Which are significantly faster, at the expense of making them less easy to read. The really big speed advantage is that you can jit this further with no python mode
`
gmt_func = layout.gmt_func
adjoint_func = layout.adjoint_func
@numba.njit
def apply_rotor_val(R_val, mv_val):
    return gmt_func( R_val, gmt_func( mv_val, adjoint_func(R_val)))
def apply_rotor(R,mv):
    return cf.MultiVector(layout,apply_rotor_val(R.value,mv.value))
`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_install:
 
 install:
   - if [[ "${CONDA}" == "true" ]]; then
-      PYTHON_VERSION="$TRAVIS_PYTHON_VERSION" CONDA_INSTALLER_OS="${TRAVIS_OS_NAME:-linux}" source auto_version/travis_install_conda.sh future numpy scipy  pip nose;
+      PYTHON_VERSION="$TRAVIS_PYTHON_VERSION" CONDA_INSTALLER_OS="${TRAVIS_OS_NAME:-linux}" source auto_version/travis_install_conda.sh future numpy scipy numba pip nose;
     fi
   - python setup.py install
 
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -238,6 +238,15 @@ def modify_idx(idx, grade):
         if not _containsDups(idx):
             done = 1
 
+def get_adjoint_function(gradeList):
+    grades = np.array(gradeList)
+    signs = np.power(-1, grades*(grades-1)/2)
+    @numba.njit
+    def adjoint_func(value):
+        return signs * value  # elementwise multiplication
+    return adjoint_func
+
+
 def get_mult_function(mult_table,n_dims):
     ''' 
     Returns a function that implements the mult_table on two input multivectors
@@ -397,6 +406,7 @@ class Layout(object):
 
         self._genEvenOdd()
         self._genTables()
+        self.adjoint_func = get_adjoint_function(self.gradeList)
 
     def __repr__(self):
         s = ("Layout(%r, %r, firstIdx=%r, names=%r)" % (
@@ -1012,13 +1022,7 @@ class MultiVector(object):
         adjoint() --> MultiVector
         """
         # The multivector created by reversing all multiplications
-
-        grades = np.array(self.layout.gradeList)
-        signs = np.power(-1, grades*(grades-1)/2)
-
-        newValue = signs * self.value  # elementwise multiplication
-
-        return self._newMV(newValue)
+        return self._newMV(self.layout.adjoint_func(self.value))
 
     __invert__ = adjoint
 
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -239,6 +239,9 @@ def modify_idx(idx, grade):
             done = 1
 
 def get_adjoint_function(gradeList):
+    '''
+    This function returns a fast jitted adjoint function
+    '''
     grades = np.array(gradeList)
     signs = np.power(-1, grades*(grades-1)/2)
     @numba.njit
--- a/docs/PerformanceCliffordTutorial.ipynb
+++ b/docs/PerformanceCliffordTutorial.ipynb
@@ -1,0 +1,785 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Writing high(ish) performance code with Clifford and Numba via Numpy\n",
+    "This document describes how to take algorithms developed in the clifford package with notation that is close to the maths and convert it into numerically efficient and fast running code. To do this we will expose the underlying representation of multivector as a numpy array of canonical basis vector coeficients and operate directly on these arrays in a manner that is conducive to JIT compilation with numba."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## First import the Clifford library as well as numpy and numba"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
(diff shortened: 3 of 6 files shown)


arsenovic:
hugo, do you know why the tests are failing? 
 `libllvmlite.so: cannot open shared object file: No such file or directory`
is this related to jitting?

its hard to read your example code in the commit, but this idea looks good to me.  soon we will have to add a page on *performance computing with clifford* to the docs, which highlight all the features you have added.

hugohadfield:
Hey Alex, just added a ipython notebook that should act as a bit of a tutorial on the jitting stuff I have been up to, it would be great to get some feedback on it as it is still in a very early stage but hopefully will be useful to people
will have a dig into why these tests are failing...

arsenovic:
hi,  
 i just took a look at you notebook. i think it is well written, and clearly demonstrates your approach. you might  comment out the prun magics ( or change the number of iterations so it doesnt take 50s!). this will also allow  it will compile on the rtd, instead of timing out.

hugohadfield:
@arsenovic the notebook now builds on rtfd and travis, are you happy for me to merge this into master?

arsenovic:
yes please do. 
 
idea. i dont want to hack this code too much to make it fast, when we know the bitmap repr is the way to go, but what do you think about having some interfaces to your improvements? like a global toggle-able 'sloppy mode'  in which the `checkother`  functions are skipped altogether?  and/or maybe there is a method  `M.rot(x)`, which nominally is  `== M*x*M.inv()` , but if MV is a even-grade multivector, then you can do your sparse speedup?  these both would essentially make your improvements more accessible.

hugohadfield:
Agreed, I think a sloppy mode would certainly be an interesting toggle, we could get some nice speedups for the average user with very minimal effort on their part

With the `M.rot(x)` idea I am not sure what the best thing to do is. Definitely adding the sparse version of the `get_mult_function` as the default in the library with it defaulting to generating a function over all grades would be worthwile and would allow easier use of the sparse stuff.
I currently have private repo of a load of fast jitted algorithms, as well as their idiomatic expressions, organised by the algebra signature that they use and some example apps, computer vision, inverse kinematics etc. I am pretty keen to open source this and would be keen to move it into pygae, but need to check with other people here re. what research they want to make public etc.

re. spending time on the making the library fast I did a little experiment
On my lab desktop for two 32 dim MVs from g3c
```
for i in range(10000000):
    layout.gmt_func(a.value,b.value)
```
takes  24.6 s

Next using the gaigen2 generated conformal code from the raytracer on Leo Dorst's website we can run the same benchmark (there is no documentation so this might be wrong, but im fairly certain its ok)
```
#include "c3ga.h"
#include <stdint.h>
using namespace c3ga;
int main(){
    mv x((unsigned int) 63,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f);
    mv y((unsigned int) 63,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f,1.0f,1.0f,1.0f,
        1.0f,1.0f);
    mv z;
    for(uint64_t i = 0; i < 10000000; i++){
        z = gp(x,y);
    }
    return 0;
}
```
This takes 9.203s

So we are 2.67 times slower than a pure c++ grade tracking bitmap implementation when doing geometric products on dense multivectors, not bad! I haven't looked at the speed difference with sparse MVs yet, need to dig into the gaigen generated code to understand how to set that benchmark up I think

hugohadfield:
So the merge looks like it has broken the docs which is very strange, they build fine on my branch but not on master

arsenovic:
all very interesting. i am a big fan of docs, and would be interested to see what you have done. 

about RTD. weird since your branch's docs built. 
i  have had problems on rtd with kernel selection in the past. if the kernel saved in the NB isnt availble on server, the sphinx-notebook fails.   i dont know if this is related, but i triggered another build.

arsenovic:
works now, not sure why
http://clifford.readthedocs.io/en/latest/PerformanceCliffordTutorial.html

hugohadfield:
@arsenovic re. sloppy mode check out 
https://github.com/pygae/clifford/blob/hugo/sloppy_mode/clifford/__init__.py

Defining a global sloppy mode that disables checking behavior and that you can trigger with a decorator:

```
_sloppy_mode = False    # global sloppy/fast mode toggle

def set_sloppy():
    global _sloppy_mode
    _sloppy_mode = True

def clear_sloppy():
    global _sloppy_mode
    _sloppy_mode = False

def sloppy(fn):
    def ret_fn(*args,**kwargs):
        set_sloppy()
        result = fn(*args,**kwargs)
        clear_sloppy()
        return result
    return ret_fn
```

Makes things a lot faster with minimal effort
```
import clifford as cf
import numpy as np
import timeit
layout,blades = cf.Cl(4,1)
n_dims = len(layout.blades)+1

R = cf.MultiVector(layout,value=np.random.rand(n_dims))
a = cf.MultiVector(layout,value=np.random.rand(n_dims))

def apply_rotor(R,a):
    return R*a*~R

def apply_rotor_fast(R,a):
    return cf.MultiVector(layout,layout.gmt_func(R.value,layout.gmt_func(a.value,layout.adjoint_func(R.value))))

@cf.sloppy
def apply_rotor_sloppy(R,a):
    return R*a*~R

apply_rotor(R,a)  20.2 µs ± 176 ns per loop
apply_rotor_sloppy(R,a) 9.76 µs ± 81.6 ns per loop
apply_rotor_fast(R,a) 6.62 µs ± 408 ns per loop
```

That branch is not a complete implementation, I only modified the functions that did not coerce types, but this is definitely useful and probably worth pursuing further

arsenovic:
sweet! i really like the decorator idea.

arsenovic:
hey wouldnt the sloppy mode test be easier to do in the begining of `_check_other`, to avoid reduncy? like 
```
def _checkOther(self,other,...):
    global sloppy_mode
    if sloppy_mode:
        return other    
...
```
or would this have an performance impact?   i guess i am thinking of possibly adding functions in the future which need to use  checkOther as well, but perhaps this is not an issue.

hugohadfield:
That sounds like a good idea, there is definitely a lot of redundancy the way I had it set up

arsenovic:
hugo, did you ever merge your sloppy_mode branch? are you still developing it?

hugohadfield:
Never merged it as got pulled onto a new project and have been a bit separated from the core GA stuff, seems unlikely I will have time till end of March at least. I think your method of doing the sloppy check at the start of checkOther is the way to go, if you want to push that stuff in I will happily nuke my branch with the partial implementation

arsenovic:
lets leave your branch for now, and who ever gets to it first wins!

Issue #15, opened 2017-11-20
object serialization and  IO


arsenovic:
it will likely be useful to be have the ability to serialize and  perhaps eventually transfer objects to other GA computational systems. the `__repr__()` methods of `Layout` and `Multivector` provide an outline of how to do this.

hugohadfield:
+1 for this. I like the idea of plain text ```__repr__()``` style serialisation for human readable files etc, although it will of course not be space efficient.
My JavaScript parseMultivector function from my GAOnline project might be a good starting point for a reverse mapping
https://github.com/hugohadfield/GAonline/blob/master/templates/main.html

arsenovic:
also, seems like big arrays of a single type of object would be common for large data sets. these would  require one map from scalars->MV, and then the array of data.

hugohadfield:
I think this is still an unsolved issue, I really dislike the multivector string parsing solution that I added! It would be much nicer to save a dictionary of blade coefficients in some form of vaguely standard file format such as Json or the numpy data file format

PR #20, opened 2018-01-16
Added minimal string parsing functionality

hugohadfield:
Adds the ability to do
```
mv_a == layout.parse_multivector(str(mv_a))
mv_a == cf.MultiVector(layout,string=str(mv_a))
```

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -175,6 +175,7 @@ from __future__ import print_function, unicode_literals
 from past.builtins import cmp, range
 from functools import reduce
 import sys
+import re
 
 # Standard library imports.
 import math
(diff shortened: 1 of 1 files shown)


hugohadfield:
Related to https://github.com/pygae/clifford/issues/15

arsenovic:
seems like accessing the MV.layout, and MV.value would be more direct like, `ugly()`?  however in the end, how its done  doesnt matter much.

hugohadfield:
Definitely true that the current ugly() print + eval combo is more direct. Makes me feel fairly uncomfortable 'eval'ing user input though, thought it might be nice to have the capacity to do both

arsenovic:
ok i see,  i agree with your string regex approx.

PR #21, opened 2018-01-24
changed the get_mult_function to operate with sparseness by default

hugohadfield:
Allows sparse product function generation but defaults to dense
Easier for users to write code that exploits sparseness of multivectors

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -261,7 +261,7 @@ def get_adjoint_function(gradeList):
     return adjoint_func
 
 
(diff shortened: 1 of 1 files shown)


hugohadfield:
The only issues I see with this are that it makes a section of the docs on writing fast clifford code outdated, I'll update them before merging

hugohadfield:
Going to close this as some of the sparse stuff is behaving a bit strangly in my code, will repoen when I have written some unit tests!

PR #24, opened 2018-04-17
Adding a check for exact object matching 

hugohadfield:
speeds up performance significantly for typical algorithms (roughly 2x) in my cases


arsenovic:
cool. i have some 4D rotor decomposition and logarithmic functions to add soon.

Issue #10, opened 2017-11-12
Performace


hugohadfield:
Hi Alex,

Having seen your issues raised with pyversor (https://github.com/tingelst/pyversor/issues/7) I agree entirely that something should be done to improve the python GA tooling.
@ereide and I work extensively with this library at Cambridge. From our perspective the key advantages that it offers are:
1. Syntax -> Close to the math (minus the operator precedence issues)
2. Distribution -> All the requirements of the package are part of the standard python scientific ecosystem. It does not require a complex build process and is available with pip. This massively reduces the barriers to getting started.

The disadvantage of this library:
1. Performance. As you have previously mentioned the matrix multiplication basis of the library is very inefficient and prevents it from being realistically useful in large applications.

What do you see as the future of this project? Do you intend to continue developing it performance-wise? 
If so, what do you think needs to be done? and how can I help?

arsenovic:
hi, good to hear this!

maybe we could all schedule  a google hangout? to discuss 
*  what everyone is doing
*  what they want/need
* how much resources are available. 

i spoke with @tingelst the other day and it was very helpful. there have been so much work done on  GA implementations, and  i think its good to think before we go coding into the sunset.  ill try to take notes and transcribe the conversions into github, so its on the web.

feel free to email me if you prefer,

eric-wieser:
FYI, [numpy is starting to think seriously about extensible dtypes again](https://youtu.be/fowHwlpGb34?t=19m46s), now that it has funding to support a small team full-time. In the long run, this ought to be a good match for representing GA types efficiently.

ereide:
I second @hugohadfield's comments.

arsenovic:
@eric-wieser , great presentation, thanks for posting this. Nathan's argument that it is more expensive to do nothing, than pay to do something is applicable for adopting  GA as well.  are you a user of clifford? perhaps you could be our ambassador to numpy?

@moble , has done work on  [quaternion](https://github.com/moble/quaternion) dtype package for numpy, so ill link him in here.  

a few ideas 

* i have created a [wiki doc](https://github.com/arsenovic/clifford/wiki/Geometric-Algebra-in-Python) in clifford to document what we might do. please add to it. my thought is to open the design discussion, and get experienced people to look at it.  (maybe we can convince ourselves not to do this!)

* i think a google hangout would go a long way also, perhaps everyone interested could respond with an available time to chat.  

[...]

PR #25, opened 2018-05-27
Hugo/sparse mult

hugohadfield:
Creates jitted sparse multiplication functions for mvs of specific grades
Now with added testing 
Note: Tests may take a while to run as it looks at all possible combinations of sparsity in grades of both objects

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -261,7 +261,7 @@ def get_adjoint_function(gradeList):
     return adjoint_func
 
 
-def get_mult_function(mult_table,n_dims):
+def get_mult_function(mult_table,n_dims,gradeList,grades_a=None,grades_b=None,filter_mask=None):
     ''' 
     Returns a function that implements the mult_table on two input multivectors
     '''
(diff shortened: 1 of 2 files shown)


arsenovic:
go right ahead. 

do you have an example of a use case for this, and the speedup it generates?

hugohadfield:
OK cool, will make a gist to explain my reasoning

hugohadfield:
OK, here is my arguments for this pull request:
https://gist.github.com/hugohadfield/f847a19aaf654f39b75f94da9b4e6fb2
Essentially it allows us to generate specialised functions for sparse multivectors in advance if we know their grade. In the case we do not specify the grades it does a runtime sparsity check as @moble suggested in https://github.com/pygae/clifford/issues/10. In the case of fully dense multivectors this will add overhead, in the more common case of sparse mvs it speeds up operations a lot for no extra effort.

hugohadfield:
I added the ability to specify the filter mask directly so as to allow highly object specific optimisation eg. for when using the pseudoscalars from various spaces or other commonly used constants where we know the exact sparsity in advance at a finer resolution than simply the grade.

PR #26, opened 2018-06-05
Jitted right inverse

hugohadfield:
This branch takes a look at the right inverse function which was previously implemented via the _myDot function. 
This PR replaces it with a jitted function that uses the sparsity of the gmt table resulting in significant speed up.
https://gist.github.com/hugohadfield/50561a0b01123e3f6e9e41f948398ee2

Additionally it adds a function that returns the modal grade of the non zero elements in a multivector. This is slightly accidentally snuck in, happy to seperate it into a different PR if that would be better.

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -187,8 +187,6 @@ from warnings import warn
 import numpy as np
 from numpy import linalg, array
 import numba
-from scipy import stats
-
 
 # The blade finding regex for parsing strings of mvs
 _blade_pattern =  "((^|\s)-?\s?\d+(\.\d+)?)\s|(-\s?(\d+((e(\+|-))|\.)?(\d+)?)\^e\d+(\s|$))|((^|\+)\s?(\d+((e(\+|-))|\.)?(\d+)?)\^e\d+(\s|$))"
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -187,7 +187,7 @@ from warnings import warn
 import numpy as np
 from numpy import linalg, array
 import numba
-
+from scipy import stats
 
 
 # The blade finding regex for parsing strings of mvs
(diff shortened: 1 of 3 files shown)


hugohadfield:
On the down side this appears to be making the python 2.7 no conda build fail on a versioning issue


TypingError: Failed at nopython (nopython frontend)
Invalid usage of Function(<unbound method OverloadTemplate_solve.solve>) with parameters (array(float64, 2d, C), readonly array(float64, 1d, C))
 * parameterized
In definition 0:
    ImportError: scipy 0.16+ is required for linear algebra
[1] During: resolving callee type: Function(<unbound method OverloadTemplate_solve.solve>)****

hugohadfield:
This PR also does not touch the other side, the left inverse, which could probably also do with a jitting!

hugohadfield:
Ok looks like the build error traces back to these lines:

``def ensure_blas():
    try:
        import scipy.linalg.cython_blas
    except ImportError:
        raise ImportError("scipy 0.16+ is required for linear algebra")


def ensure_lapack():
    try:
        import scipy.linalg.cython_lapack
    except ImportError:
raise ImportError("scipy 0.16+ is required for linear algebra")
``

Not really sure what to do about this, on the one hand I would really like to have a fast inverse in the library, on the other I don't really like the idea of forcing users to have super up to date libraries to run the code. I could separate them into two functions, the slower old one that operates by default and leave the optimised version attached to Layout in case people want to use it? @arsenovic what do you think?

arsenovic:
i dont understand why there are 3 inverse functions, in the first place. 

i looked up scipy0.16 and its 3 yrs old, so i think its a reasonable requirement. however, i like your general idea of kindly supporting old versions, as long as it doesnt make the code too unclear or annoying for you. so i say its up to you.

hugohadfield:
I also am not sure yet why there are three! Will probably have a dig through how they work/why there are multiple/where we can speed them up if we know the sparsity of the output objects in advance over the next few weeks...
I'm not sure how library versioning works for travis, will have a dig into it and see if we can get it to build with updated versions for the non conda case, I think 3 years old is definitely reasonable!

hugohadfield:
boom, adding scipy to the requirements lets it build

PR #35, opened 2018-07-18
Oo cga

arsenovic:
playing around with a more object oriented approach to cga,  

here is a shelled out  [demo]( https://gist.github.com/arsenovic/f23c696c17f3ee7de529028d0cfcc085), supporting only spheres and translations. the goal is to  make the code very concise, and prevent  redundancy for things based on specific dimensions. so g2c and g3c should work very similarly.

@hugohadfield , let me know what you think about this approach. 
i am  not intending to merge yet. (and i didnt intend to submit the caching with this, accident)


hugohadfield:
So I think there are some big benefits to be had from using an object oriented approach with conformal, it can improve usability and we can exploit properties of typed multivectors to make things faster etc.

Personally however I feel that I would not use it much. For me one of the main attractions to this library was the similarity to the maths and using untyped multivectors gives me the flexibility to play around however I choose without having to fix anything as a specific object.

hugohadfield:
Sounds good, ill look at the conflicts now!

arsenovic:
thats my bug

arsenovic:
well, i suppose you would have to typecast it to a MV, or use  layout.scalar*0. 
or just try/except and avoid the test of x.layout.

PR #42, opened 2018-08-17
RFC: Speed up the _genEvenOdd calculations

moble:
This isn't necessarily meant to be merged, as it might be too much code bloat for not enough benefit, but it grew out of [a comment I made on a commit of Hugo's](https://github.com/pygae/clifford/commit/f67bcefa460560b8317583601e49b5631b556a0c#commitcomment-30107076), and I think it would be nice to preserve it somewhere more standard.

The basic idea is that when you loop over all permutations, at least using `itertools.permutations`, it's easy to compute the sign just based on the index.  Specifically, when the permutation follows "lexicographic" ordering, if you're on permutation `i`, then the sign of the permutation is
```python
(-1)**sum(i//factorial(j) for j in range(0, total_number_of_permutations, 2))
```
Plus, we always know the sign of every second permutation based on the preceding sign, so we can skip the calculation of half of the signs by just grabbing the next iteration and doing what we know we should.

It turns out that compiling the code to do the sum above is time consuming, and even numpy isn't too good at this sort of thing, so the best solution for reasonable dimensions (here I've chosen 16) is to just hard-code all of the terms in the sum.

Here are the timings comparing the original code to this new code when initializing algebras of the given dimensions:

| Dimension | Original (s) | New (s) |
|-----------|----------|-----|
| 2 | 0.0017871856689453125 | 0.0014159679412841797 |
| 3 | 0.00167083740234375 | 0.0016551017761230469 |
| 4 | 0.004223823547363281 | 0.004168033599853516 |
| 5 | 0.01593494415283203 | 0.015892982482910156 |
| 6 | 0.08282613754272461 | 0.07908296585083008 |
| 7 | 0.4307372570037842 | 0.40534186363220215 |
| 8 | 2.685001850128174 | 2.4293580055236816 |
| 9 | 19.3399920463562 | 16.702773809432983 |
| 10| 220.34591484069824 | 192.91048669815063 |

Note that while I have checked that this `sign` function agrees with the original `_sign` function, I haven't checked that this code actually returns the same results as before.

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -202,22 +202,6 @@ def get_longest_string(string_array):
     """
     return max(string_array,key=len)
 
-def _sign(seq, orig):
-    """Determine {even,odd}-ness of permutation seq or orig.
-
-    Returns 1 if even; -1 if odd.
-    """
-
-    sign = 1
-    seq = list(seq)
-
-    for i in range(len(seq)):
-            if seq[i] != orig[i]:
-                j = seq.index(orig[i])
-                sign = -sign
-                seq[i], seq[j] = seq[j], seq[i]
-    return sign
-
 
 def get_adjoint_function(gradeList):
     '''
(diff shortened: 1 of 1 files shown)


moble:
Closing this as irrelevant, given Hugo's excellent work #43.

PR #43, opened 2018-08-17
Remove large sparse matrices from layout object

hugohadfield:
This commit removes the very large and very sparse matrices from the Layout object and instead directly generates a dictionary of values that would index into the sparse matrix.
This results in much lower memory  usage and is a step towards fast and efficient caching of algebras, we should just need to store these sparse objects.
In order to get this to work I had to nuke the right inverse function. I am not entirely sure why this was in there to begin with and its removal does not seem to break any of my code...

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -203,23 +203,6 @@ def get_longest_string(string_array):
     return max(string_array,key=len)
 
 
-def _sign(seq, orig):
-    """Determine {even,odd}-ness of permutation seq or orig.
-
-    Returns 1 if even; -1 if odd.
-    """
-
-    sign = 1
-    seq = list(seq)
-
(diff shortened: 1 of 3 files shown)


hugohadfield:
My latest commit brings the start up time for high dimensional algebras down significantly
On my machine:

2 0.1275029182434082
3 0.0009195804595947266
4 0.0021293163299560547
5 0.006368160247802734
6 0.023430585861206055
7 0.09814119338989258
8 0.41788220405578613
9 1.6764190196990967
10 6.97962498664856

It implements the bitmap style operations described in chapter 19 of Leo Dorsts book

hugohadfield:
@Stefan-Endres you will probably be interested in this, should make your application much more feasible

arsenovic:
sweet!

PR #47, opened 2018-08-22
Testing rebuild

hugohadfield:
Fixes firstIdx bug
Improves testing setup to operate on all files in a test directory
Adds a check for cuda to allow cuda testing if available


hugohadfield:
This breaks the tests as the new space time test fails on python2.7. I think it is still worth merging to master as I don't think it is actually breaking anything, just highlighting an existing error

PR #56, opened 2018-10-03
Generation optimise

hugohadfield:
Large optimisations to the algebra generator, should be **much** faster for high dimensional algebras now


hugohadfield:
The main advantages here come from JITing of the algebra generation and automatic parallelisation. Some timings on my 8 core machine:
dimension | time(s)
2 | 0.6529016494750977
3 | 0.0006191730499267578
4 | 0.0006620883941650391
5 | 0.00083160400390625
6 | 0.0013201236724853516
7 | 0.00405430793762207
8 | 0.012106895446777344
9 | 0.041641950607299805
10 | 0.17168784141540527
11 | 0.7451441287994385

hugohadfield:
Now even faster

dimension | time(s)
--- | --- 
2 | 0.6636462211608887
3 | 0.0006034374237060547
4 | 0.0006799697875976562
5 | 0.0008611679077148438
6 | 0.0013630390167236328
7 | 0.00307464599609375
8 | 0.010649919509887695
9 | 0.03726935386657715
10 | 0.14243125915527344
11 | 0.5639963150024414

arsenovic:
why is dim 2 slower ?

arsenovic:
(and wow thats really fast. )

hugohadfield:
Dim 2 is slower probably because it contains the runtime of the jit compiler, after the functions are jitted once for the correct function signature it just uses them for all the other cases without having to recompile :)

hugohadfield:
And yeah it's super quick, made me wonder for a long time if it was doing something wrong, but passes all the tests and I in theory it is the same algorithm

hugohadfield:
Still eats a lot of memory high dim algebras, makes me wonder about making a little spinoff version optimised with cuda that I can run on some very high memory machines here at the uni and we just generate the sparse objects on those and upload them to some repo. Then any future ga implementation can just pull in the arrays from online

arsenovic:
i like it.  so we are getting back into the algebra caching idea, except you want to cache it in the cloud for algebras too big to generate locally .

PR #62, opened 2018-10-25
Code generator

hugohadfield:
Upgrades tools related to CUDA implementation
Also accidentally includes a test hdf5 serialisation implementation


clifford/__init__.py
```diff
@@ -53,6 +53,9 @@
 from numpy import linalg, array,zeros
 import numba
 
+from clifford.gafileio import write_ga_file, read_ga_file
```
arsenovic: i would have just called this `io`

hugohadfield: Agreed, I think that is nicer!

hugohadfield: Will fix it with some other improvements to the tooling today :)

PR #65, opened 2018-11-02
Replace line and plane intersection function with njited one

sushachawal:
Replaced the old function to test the intersection between a plane and a line and return the point with a new one. The name of the original function is kept the same for ease

### DIFF
--- a/clifford/tools/g3c/__init__.py
+++ b/clifford/tools/g3c/__init__.py
@@ -443,11 +443,11 @@ def val_intersect_line_and_plane_to_point(line_val, plane_val):
     Returns the point at the intersection of a line and plane
     """
     m = meet_val(line_val, plane_val)
-    if(gmt_func(m,m)[0] < 0.000001):
+    if gmt_func(m, m)[0] < 0.000001:
         return np.array([-1.])
     output = np.zeros(32)
     A = val_normalised(m)
-    if(A[15] < 0):
+    if A[15] < 0:
         output[1] = A[8]
         output[2] = A[11]
         output[3] = A[14]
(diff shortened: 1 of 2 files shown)


hugohadfield:
Looks like it needs a sign fox in a few cases, maybe to do with the orientation of the line vs orientation of the plane? Check out the sign fix that was in there originally!

sushachawal:
The sign fix is done now! Basically we need the following fact to be true:
`meet(L, P) == meet(-L, P)` since the orientation of the line or plane does not affect the location of the point of intersection. This sign however is **carried through into the meet** so to un-sign, we use the following analysis:

<a href="https://www.codecogs.com/eqnedit.php?latex=\begin{align*}&space;\phi&space;\vee&space;L&space;=&space;X&space;\wedge&space;n&space;=&space;x&space;\wedge&space;n&space;&plus;&space;n&space;\wedge&space;\bar&space;n&space;=&space;x&space;\wedge&space;n&space;&-e_{45}\\&space;\phi&space;\vee&space;L&space;=&space;-X&space;\wedge&space;n&space;=&space;-x&space;\wedge&space;n&space;-&space;n&space;\wedge&space;\bar&space;n&space;=&space;-x&space;\wedge&space;n&space;&&plus;&space;e_{45}&space;\end{align*}" target="_blank"></a>

Notice that if we extracted without checking for the sign on e_45 we would actually return -x!

The unit tests in Travis CI fail due to a precision error in the underlying value vector. The new val_ function has float precision whereas the unit test was expecting integer precision.

hugohadfield:
@sushachawal any chance you could fix the test to match the function and add to it to your PR?

sushachawal:
Yeah sure, will add it later today or tomorrow morning :)

sushachawal:
The function now returns a conformal point as opposed to the previous version which returned a 3D point. The unit test has now been fixed to account for this.

PR #66, opened 2018-11-14
Tool improvements

hugohadfield:
Adds some miscellaneous conformal tools 
Adds the ability to write json scenes for importing into ganja.js
Adds the ability to write json in the same format as the hdf5 files


clifford/tools/g3c/GAOnline.py
```diff
@@ -175,9 +176,64 @@ def add_object_array(self,mv_array,mv_type='interp',color='rgb(0,0,0)'):
...
             self.add_object(mv, mv_type, color=color)
 
-    def save_to_file(self,filename):
+    def save_to_file(self, filename):
         with open(filename,'w') as fobj:
-            print(self,file=fobj)
+            print(self, file=fobj)
+
+
+class GanjaScene:
```
arsenovic: cool idea

hugohadfield: the plan is to knock up a proper interface to let us generate iframes etc for display in jupyter notebooks/separate window, im keen for a matplotlib but for ga :)

PR #71, opened 2018-12-07
Fixes for exponential, drop python2

hugohadfield:


hugohadfield:
@arsenovic we can ditch this right? Fairly certain we got the useful stuff in the otehr PR right?

PR #87, opened 2019-03-23
add sta as a premade algebra

arsenovic:


arsenovic:
fails are unrelated to this pr

PR #88, opened 2019-03-26
added array() method to make MVarray use easier, removed numpy array …

arsenovic:
…import from init.py. @hugohadfield  do you think re-using `array()` like this is bad?  i figured it would only cause problems if you do `from clifford import * ` and it overwrites `numpy.array`

this now works, which is a little easier 

    from clifford import g3,array
    import numpy as np
    theta = np.linspace(0,2*np.pi,10)
    np.e**(theta*array(g3.e12))

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -50,7 +50,7 @@ from warnings import warn
 
 # Major library imports.
 import numpy as np
-from numpy import linalg, array,zeros
+from numpy import linalg,zeros
 import numba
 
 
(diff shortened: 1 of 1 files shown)


hugohadfield:
Yeah I like it, I wonder if we could also hijack the array wrap stuff in here:
https://github.com/pygae/clifford/blob/19bfa2d42d81df89b57c0b4bbd307b2bcb70c27b/clifford/__init__.py#L981
To try and do it even without the array method...

hugohadfield:
Lets merge this in the meantime and have a think, from conversing with the Czech guys I think more array methods are definitely wanted by users

PR #89, opened 2019-03-27
Potential change to allow array ops with MultiVectors

hugohadfield:
@arsenovic check this one out and have a play, needs a lot more tests but seems to have minimal functionality...

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -965,6 +965,7 @@ class MultiVector(object):
         """
 
         self.layout = layout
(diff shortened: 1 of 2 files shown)


arsenovic:
hugo this works great, its exactly what i wanted, and your tests are good. 

the implementation is what i was trying to avoid because of the high level of redundancy, which seems to me it could/should be avoided.  

regardless i making it work is more important than making the implementation clean.  so this is a big plus for clifford.

Issue #6, opened 2016-10-19
array of multivectors fails


arsenovic:
@rkern or @moble,  do either of you know why  this happens? 

```
import numpy as np
from clifford import Cl 
layout,blades = Cl(2) # note i changed firstIdx=1 by default in Cl
locals().update(blades)

np.array([e1,e2],dtype=np.object) 
```

yields

```
array([[0, 1, 0, 0],
       [0, 0, 1, 0]], dtype=object)
```

perhaps there are some special array-like methods that `Multivector` implements? is there a way to have an array of multivectors, instead of an array of their `value`s?

rkern:
`Multivector` objects are also sequences. `np.array()` consumes sequences as if they were lists. To create a `dtype=object` array of sequences, the magic of `np.array()` that automatically determines the shape of the output is interfering. The usual workaround is to create an empty `object` array of the desired shape, then assign the elements.

```
vs = [e1, e2]
a = np.empty(len(vs), dtype=object)
a[:] = vs
```

arsenovic:
sweet thanks a lot.    

fyi, i am currently working on  a [ `Frame` ] (https://github.com/arsenovic/clifford/blob/master/clifford/__init__.py#L1639) object. its a light object, and i was looking for  simple element-wise arithmetic  to work, like  i could implement `A*B`  with 

```[a*b for a,b in zip(A,B)]```

but then  it seemed that i should  just to re-use numpy's array for this. However,  when i make two np.array's of clifford vectors, and try A*B it doesnt produce the expected result result...

**Comment by arsenovic** [comment]
i am re-opening. because clifford is numerical package, we need to make working with arrays easier, so that  broadcasting  works cleaner. 

ie i want something like this to work. 

    from clifford.g3 import * 
    x = linspace(0,1,101)*e3

**Comment by arsenovic** [comment]
i have tried a few things with `__array_ufunc__`, and `__array_wrap__` , but the difficulty of this problem doesnt make sense to me.  all that is needed is for `ndarray`'s to see `MultiVector`s  as numbers instead of sequences,  and  i think all the broadcasting and ufuncs(that are defined) should work?

**Comment by hugohadfield** [comment]
@eric-wieser you probably know all about this stuff, any chance of a little help :) ?
We are basically aiming to be able to write something like:
```
(np.random.rand(4)*e1 )+e12
```
or 
```
e12+(e1*np.random.rand(4))
```
We currently have an array wrap setup here:
[...]

PR #90, opened 2019-04-11
Alternate array setup

hugohadfield:
Uses Eric's suggestions from here: https://github.com/pygae/clifford/issues/6
Removes array wrap
@arsenovic check it out

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -979,44 +979,9 @@ class MultiVector(object):
                     "value must be a sequence of length %s" %
                     self.layout.gaDims)
 
-    def __array_wrap__(self, out_arr, context=None):
-        '''
-        This is a work-around needed to prevent numpy arrays from
-        operating  on MultiVectors using their operators.  This happens
-        because Multivectors act enough like an array, that duck
-        typing thinks they are an array.
-        '''
-        
-        uf, objs, out_index = context
-        op0 = objs[0]
-
-        # If its an exp it only has one argument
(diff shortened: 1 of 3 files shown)


arsenovic:
looks right

clifford/__init__.py
```diff
@@ -979,44 +979,9 @@ def __init__(self, layout, value=None, string=None):
...
-        elif uf.__name__ == 'add':
-            return op1 + op0
-        elif uf.__name__ == 'subtract':
-            return -op1 + op0
-
-        else:
-            raise ValueError('i dont know how to %s yet' % uf.__name__)
+    def __array__(self):
+        # we are a scalar, and the only appropriate dtype is an object array
+        return MVArray([self])
```
eric-wieser: This is somewhat sloppy, ideally you'd return a 0d array here with `MVArray(self)`. Looking at `MVArray` though, that doesn't look supported right now.

clifford/__init__.py
```diff
@@ -1067,8 +1035,7 @@ def __mul__(self, other):
             newValue = self.layout.gmt_func(self.value,other.value)
         else:
             if issubclass(other.__class__, np.ndarray):
-                obj = np.empty(1, dtype=object)
-                obj[0] = self
+                obj = self.__array__()
```
eric-wieser: `obj = np.asanyarray(self)` might be more idiomatic - usually you shouldn't invoke `__dunder__` methods directly

PR #99, opened 2019-10-06
Fix Round dimension

rotu:
Round dimension was off by 2 - the extra dimensions from the representation space.


rotu:
I'm not sure if this is correct, due to my poor understanding of GA, but I assume that `CGA(m).round(n).dim` should be `n`

PR #100, opened 2019-10-06
Fix silly mistake in Flat constructor

rotu:
Previously, checking for membership in map object would exhaust the generator and result in a trivial Flat.

### DIFF
--- a/clifford/cga.py
+++ b/clifford/cga.py
@@ -164,7 +164,7 @@ class Flat(CGAThing):
             
         # from vectors on flat         
         else:
-            nulls = map(self.cga.null_vector,args)
+            nulls = list(map(self.cga.null_vector,args))
             if self.einf not in nulls:
                 nulls = list(nulls)+[self.einf]
                 


clifford/cga.py
```diff
@@ -165,10 +165,8 @@ def __init__(self,cga, *args):
         # from vectors on flat         
         else:
             nulls = list(map(self.cga.null_vector,args))
-            if self.einf not in nulls:
-                nulls = list(nulls)+[self.einf]
-                
-            self.mv = reduce(op,nulls) 
+            self.mv = reduce(op,nulls)
+            self.mv = op(self.mv, self.einf) or self.einf
```
arsenovic: can you explain more about this? i dont follow  it.

rotu: Oops. I made my own silly mistake. It should be `op(self.mv, self.einf) or self.mv`. If a multiple of einf is in the list of blades, it will outer to zero, and we shouldn’t multiply in that additional einf factor.

arsenovic: why not

clifford/cga.py
```diff
@@ -164,11 +164,9 @@ def __init__(self,cga, *args):
             
         # from vectors on flat         
         else:
-            nulls = map(self.cga.null_vector,args)
-            if self.einf not in nulls:
-                nulls = list(nulls)+[self.einf]
-                
-            self.mv = reduce(op,nulls) 
+            nulls = list(map(self.cga.null_vector,args))
```
eric-wieser: Is it sufficient to change only this line? What is the effect of changing the other lines? Is it possible that `op(self.mv, self.einf)  == 0` but `self.einf not in nulls`?

eric-wieser:
Did some more reading, your changes look obviously correct to me.

Can you add a test?

clifford/cga.py
eric-wieser: Makes sense to me, thanks for the summary.

PR #106, opened 2019-10-11
Remove the nose dependency

### LINKED ISSUE #105: Switch to pytest
From the docs for nose:

> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2.

Pytest seems to be the standard in the scientific python ecosystem

eric-wieser:
This module is abandoned, and isn't really being used anyway.

Proof that this works will be in CI somewhat-passing

Closes #105

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,13 +23,13 @@ before_install:
 
 install:
   - if [[ "${CONDA}" == "true" ]]; then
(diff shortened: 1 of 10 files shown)


eric-wieser:
CI is working to the extent that it currently works.

PR #107, opened 2019-10-11
Fix DeprecationWarning: invalid escape sequence

eric-wieser:
If `\` is not used as part of an escape sequence, it needs to be either a raw string or escaped as `\\`.

This is needed to work at all in python 3.8


eric-wieser:
Once again, CI failures are unrelated

PR #108, opened 2019-10-11
Fix warnings from LGTM

eric-wieser:
Found via https://lgtm.com/projects/g/pygae/clifford


eric-wieser:
Unusually, Travis CI actually passed!

Will go ahead and merge...

PR #109, opened 2019-10-11
Add PyPI classifiers and python_requires

eric-wieser:


eric-wieser:
Merging early, there's nothing interesting to test here - all CI runs got past the install step

PR #110, opened 2019-10-11
Remove remnants of python 2 support

eric-wieser:


eric-wieser:
@hugohadfield: Any idea why azure isn't running here?

hugohadfield:
I have no idea..

Issue #104, opened 2019-10-10
Travis builds are failing


eric-wieser:
Excerpt [from these lines](https://travis-ci.org/pygae/clifford/jobs/596237082#L468-L499):
```
======================================================================
FAIL: test_motor_between_rounds (clifford.test.transplant_class.<locals>.C)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/pygae/clifford/clifford/test/test_g3c_tools.py", line 635, in test_motor_between_rounds
    np.testing.assert_almost_equal(C2.value, C3.value, 3)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/numpy/testing/_private/utils.py", line 568, in assert_almost_equal
    return assert_array_almost_equal(actual, desired, decimal, err_msg)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/numpy/testing/_private/utils.py", line 973, in assert_array_almost_equal
    precision=decimal)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/numpy/testing/_private/utils.py", line 789, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not almost equal to 3 decimals
(mismatch 6.25%)
 x: array([-0.   ,  0.   ,  0.   ,  0.   ,  0.   ,  0.   ,  0.   ,  0.   ,
       -0.   , -0.   ,  0.   , -0.   , -0.   , -0.   , -0.   , -0.   ,
        0.   ,  0.   ,  0.   ,  0.   ,  0.   ,  0.   ,  0.   ,  0.   ,
        0.   ,  0.   , 20.374, 20.374,  0.862, -0.47 ,  0.19 ,  0.   ])
 y: array([ 0.   ,  0.   ,  0.   ,  0.   ,  0.   ,  0.   , -0.   ,  0.   ,
        0.   ,  0.   ,  0.   , -0.   , -0.   , -0.   , -0.   ,  0.   ,
        0.   ,  0.   ,  0.   ,  0.   ,  0.   ,  0.   ,  0.   ,  0.   ,
        0.   ,  0.   , 20.542, 20.542,  0.862, -0.47 ,  0.19 ,  0.   ])
-------------------- >> begin captured stdout << ---------------------
random_point_pair 2
random_circle 3
random_sphere 4
[...]

PR #111, opened 2019-10-11
Use pytest to parametrize tests

eric-wieser:
Changes involved here:

* Tidying imports
* Replacing in-test for loops with parametrization of:
  * algebras
  * object generators
* Replacing unittest.TestCase objects with regular classes, so that they work with pytest (and renaming from XTests to TestX)
* Merging similar tests into a parametrized test

Easier to review with

image


eric-wieser:
Merging in the hope this will unblock #104

PR #112, opened 2019-10-11
Mark the worst offender of the flaky tests as xfail

eric-wieser:
This will prevent CI runs failing due to it

#104


eric-wieser:
Worked as intended:

https://travis-ci.org/pygae/clifford/jobs/596662145#L1018

PR #114, opened 2019-10-14
Parametrize another test

eric-wieser:


eric-wieser:
Breakage is real, I can't spell `paramterize`

eric-wieser:
Passes at least once on travis, suggesting we're back to non-deterministic failures rather than typos.

New parametrized tests are showing up in azure

PR #117, opened 2019-10-14
Various cleanups to get_mult_function

eric-wieser:
Again, I would recommend this be reviewed commit by commit

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -160,78 +160,50 @@ def get_mult_function(k_list, l_list, m_list, mult_table_vals, n_dims, gradeList
     '''
     Returns a function that implements the mult_table on two input multivectors
     '''
-    if product_mask is None:
-        k_list_copy = k_list
-        l_list_copy = l_list
-        m_list_copy = m_list
-        mult_table_vals_copy = mult_table_vals
-    else:
-        # We can pass the sparse filter mask directly
-        k_list_copy = np.zeros(product_mask.shape[0], dtype=np.int64)
-        l_list_copy = np.zeros(product_mask.shape[0], dtype=np.int64)
-        m_list_copy = np.zeros(product_mask.shape[0], dtype=np.int64)
-        mult_table_vals_copy = np.zeros(product_mask.shape[0])
-        for i in range(product_mask.shape[0]):
-            k_list_copy[i] = k_list[product_mask[i]]
-            l_list_copy[i] = l_list[product_mask[i]]
-            m_list_copy[i] = m_list[product_mask[i]]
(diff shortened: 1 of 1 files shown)


clifford/__init__.py
```diff
@@ -155,87 +157,101 @@ def construct_tables(
...
             if v_val != 0.0:
-                m = m_list_copy[ind]
+                m = m_list[ind]
                 ov_val = other_value[m]
                 if ov_val != 0.0:
-                    l = l_list_copy[ind]
-                    output[l] += v_val * mult_table_vals_copy[ind] * ov_val
+                    l = l_list[ind]
+                    output[l] += v_val * mult_table_vals[ind] * ov_val
         return output
```
eric-wieser: @hugohadfield: Does this actually help vs just doing the multiply by zero?

If not, this entire function can be removed.

eric-wieser:
Last commit (f3a5a5d4281c707274764a937b2fce0b4d1af344) made travis noticeably slower, so backing it out for now

clifford/__init__.py
```diff
@@ -160,77 +160,91 @@ def get_mult_function(k_list, l_list, m_list, mult_table_vals, n_dims, gradeList
...
-        for i in range(product_mask.shape[0]):
-            k_list_copy[i] = k_list[product_mask[i]]
-            l_list_copy[i] = l_list[product_mask[i]]
-            m_list_copy[i] = m_list[product_mask[i]]
-            mult_table_vals_copy[i] = mult_table_vals[product_mask[i]]
+    if product_mask is not None:
+        k_list = k_list[product_mask]
+        l_list = l_list[product_mask]
+        m_list = m_list[product_mask]
+        mult_table_vals = mult_table_vals[product_mask]
```
eric-wieser: Note that this loop is just a regular numpy indexing operation

clifford/__init__.py
```diff
@@ -160,77 +160,91 @@ def get_mult_function(k_list, l_list, m_list, mult_table_vals, n_dims, gradeList
...
+    @numba.njit
+    def mv_mult(value, other_value):
+        output = np.zeros(n_dims)
+        for ind, k in enumerate(k_list):
+            m = m_list[ind]
+            l = l_list[ind]
+            output[l] += value[k] * mult_table_vals[ind] * other_value[m]
+        return output
+
+    return mv_mult
```
eric-wieser: This code just moved

clifford/__init__.py
hugohadfield: The first check almost definitely should (if v_val != 0.0:) but the second im not so sure. Once upon a time I did make a notebook comparing these things.. I think it is worth setting up a proper benchmarking system

PR #120, opened 2019-10-15
Remove NoMorePermutations

eric-wieser:


eric-wieser:
Code that isn't there won't be broken, merging...

PR #125, opened 2019-10-15
Rewrite clifford.elements to use itertools

eric-wieser:
This function is really just the `powerset` reciple from the itertools docs.

Timings:
```python
>>> %timeit clifford.elements(3)
83.7 µs ± 1.82 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

>>> %timeit list(powerset(range(3)))
5.09 µs ± 424 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

>>> %timeit list(powerset(range(12)))
313 µs ± 14.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

>>> %timeit clifford.elements(12)
11.2 ms ± 3.94 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

>>> assert list(powerset(range(12))) == clifford.elements(12)
```

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -2273,71 +2273,22 @@ class BladeMap(object):
         return B
 
 
-def comb(n, k):
-    """\
-    Returns /n\\
-            \\k/
-
-    comb(n, k) --> PyInt
-    """
-
-    def fact(n):
-        if n == 0:
-            return 1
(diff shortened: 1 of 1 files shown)


eric-wieser:
Travis passes, local tests prove that this is faster and equivalent - merging

PR #126, opened 2019-10-15
Enable codecov on travis

eric-wieser:
Probably not going to work first time...

Could also consider using azure by itself for this: https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/python?view=azure-devops#test-with-pytest-and-collect-coverage-metrics-with-pytest-cov

This makes the tests take about 50% longer, but I think that's an ok cost. We might be able to shave off some time elsewhere to make up for it - eg by recycling algebra objects.

https://codecov.io/github/pygae/clifford/commit/097ba40367938d5579af9d516c57135b2f7c4edb


eric-wieser:
Let's put this in, and see what happens

PR #127, opened 2019-10-15
Remove some duplicate code in MultiVector.__str__

eric-wieser:
Also add a unit test for it


eric-wieser:
CI failed due to #131, trying again with a merge

eric-wieser:
One of the four travis builds passed, rest are transient it seems

Issue #93, opened 2019-05-07
Errors on PerformanceCliffordTutorial


cdeil:
There's currently errors on https://clifford.readthedocs.io/en/latest/PerformanceCliffordTutorial.html 

`AttributeError: 'Layout' object has no attribute 'gmt'`

Needs a fix?

Screenshot:

![Screenshot 2019-05-07 at 18 34 21](https://user-images.githubusercontent.com/852409/57316744-cde34480-70f6-11e9-920c-eeacfefd90b9.png)

hugohadfield:
Yes it does, apologies! It is a bit out of date due to some overhauls of the internals of the library a while ago

eric-wieser:
#138 fixes this error, but introduces a new one: https://external-builds.readthedocs.io/html/clifford/138/PerformanceCliffordTutorial.html

(note this link may become stale)

![image](https://user-images.githubusercontent.com/425260/67009488-a51e9c80-f0e3-11e9-9073-d60b3b7a4a91.png)

`AttributeError: 'Layout' object has no attribute 'omt'`

eric-wieser:
Fixed in gh-140

PR #133, opened 2019-10-16
Combine the k_list, l_list, m_list, and mult_table_vals into a class

eric-wieser:
Combine the k_list, l_list, m_list, and mult_table_vals into a class …

This gives a convenient place to attach methods, and reduced the number of arguments for a lot of functions

Since this moves functions to methods, review with this setting turned on:
image

---

Add nonzero and getitem to fix gh-93

These methods are super inefficient, but they are enough to fix the old code.

---

This change is pretty wide-reaching, and I probably haven't got it right first time. This should make the API compatible with how it was before gh-93 (!).

It's likely I will try to split this into smaller PRs (done)

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -179,7 +179,7 @@ class _MultiplicationTable:
         self._val_list = val_list
         self._dims = dims
 
-    # fake being an array
+    # fake being an array, for compatibility with older code
     @property
     def ndim(self):
         return 3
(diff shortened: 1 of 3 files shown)


clifford/__init__.py
```diff
@@ -193,53 +167,194 @@ def _get_mult_function(k_list, l_list, m_list, mult_table_vals, n_dims):
...
+        # reshape the array to handle broadcasting
+        index_shape = np.broadcast(k, l, m, self._val_list).shape
+        broadcast_vals = np.broadcast_to(self._val_list, index_shape)
+
+        matching = (k == self._k_list) & (l == self._l_list) & (m == self._m_list)
+        try:
+            # numpy 1.17
+            return np.sum(broadcast_vals, where=matching, axis=-1)
+        except TypeError:
+            return np.sum(broadcast_vals*matching, axis=-1)
```
eric-wieser: gh-93 required `gmt[gmt.nonzero()]` to work - this makes that and variants behave correctly.

clifford/__init__.py
```diff
@@ -193,53 +167,194 @@ def _get_mult_function(k_list, l_list, m_list, mult_table_vals, n_dims):
...
+        return intermed
+
+    def _val_get_left_matrix(self, x):
+        """
+        This produces the matrix X that performs left multiplication with x
+        eg. X@b == (x*b).value
+        """
+        return _MultiplicationTable._numba_val_get_left_matrix(
+            self._k_list, self._l_list, self._m_list, self._val_list, self._dims
+        )
```
eric-wieser: This moved from the top of the diff

clifford/test/test_clifford.py
```diff
@@ -349,8 +349,7 @@ def test_right_multiplication_matrix(self, algebra):
         for i in range(1000):
             a = layout.randomMV()
             b = layout.randomMV()
-            b_right = val_get_right_gmt_matrix(b.value, layout.k_list, layout.l_list, layout.m_list,
-                                               layout.mult_table_vals, layout.gaDims)
+            b_right = layout.get_right_gmt_matrix(b.value)
```
eric-wieser: This cleanup worked just fine before this change too.

eric-wieser:
Closing, did #138 instead

PR #134, opened 2019-10-16
Use `layout.get_right_gmt_matrix` instead of the raw function

eric-wieser:
Split from gh-133.

Will merge if CI passes.


eric-wieser:
`clifford/test/test_clifford.py` passed in CI, let's shortcut the rest of the build

PR #135, opened 2019-10-16
Used index masking for code generator like was done for the multiply function

eric-wieser:
Split from gh-133.

Will merge if CI passes.

Code generator was rerun locally, did not crash and produced no changes.


eric-wieser:
Some CI passes, and this probably isn't run by CI anyway.

PR #138, opened 2019-10-16
Use sparse.COO for our multiplication table

eric-wieser:
An alternative to #133, that shares the first two commits

I plan to squash these commits when I merge, the history isn't that interesting

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -179,7 +179,7 @@ class _MultiplicationTable:
         self._val_list = val_list
         self._dims = dims
 
-    # fake being an array
+    # fake being an array, for compatibility with older code
     @property
     def ndim(self):
         return 3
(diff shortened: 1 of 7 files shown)


clifford/__init__.py
```diff
@@ -139,73 +142,70 @@ def construct_tables(
...
                       grades_a=None, grades_b=None, filter_mask=None):
     '''
     Returns a function that implements the mult_table on two input multivectors
     '''
     if product_mask is not None:
-        k_list = k_list[product_mask]
-        l_list = l_list[product_mask]
-        m_list = m_list[product_mask]
-        mult_table_vals = mult_table_vals[product_mask]
+        mt = sparse.where(product_mask, mt, 0)
```
eric-wieser: Strictly this is a little less efficient, but it has the advantage of working with dense product masks too. At any rate, this is only a startup cost.

hugohadfield: How does this start up impact effect large-ish d algebras eg Cl(8)?

eric-wieser: Before:
```
2.16 s ± 47.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
After:
```
In [2]: %timeit clifford.Cl(8)
1.34 s ± 64.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

???

hugohadfield: Nice

eric-wieser:
A little worried that `test_rotor_between_non_overlapping_spheres` is failing on both travis conda runs

eric-wieser:
I restarted the travis and a different test failed - looks like that was just coincidence

PR #139, opened 2019-10-17
Swap the l and m axes in the cuda generator to restore our original order

eric-wieser:
We emit data ordered by (k, m, l), but sparse.COO reorders it in memory to the order (k, l, m). Transposing brings the order back.

Tested by regenerating cuda_products.py

This should have been part of gh-138, but I forgot to run it.


eric-wieser:
Merging without CI since the code this touches isn't run by CI

PR #142, opened 2019-10-18
De-nest `isBlade` and `isVersor` for readability, and add comments from Dorst

eric-wieser:


eric-wieser:
Travis passed, merging...

PR #143, opened 2019-10-18
Speed up grades_present by a factor of 5, and change it to return a set not a list

eric-wieser:


eric-wieser:
Travis passes, without any noticeable speed change. Merging...

PR #147, opened 2019-10-18
Split clifford/__init__.py into multiple files

eric-wieser:
Process for this was:
* Extract large classes into their own files
* Run flake8 to find the missing imports
* Use a mixture of:
  * `from . import name` for `name`s which are defined before the class
  * `cf.name` for `name`s which are defined after the class
* Replace the class definition with an import of that class from the new destination

Arguments for this change:

* Less scrolling to discover relevant functions
* More granular code coverage

Arguments against:
* Makes git history harder to follow
* Leaves behind some circular imports to be cleaned up later


eric-wieser:
@utensil [remarked](https://clifforddev.slack.com/archives/CH0NA10LE/p1570793923010200):

> I'm not involved in clifford's development and not a frequent user of it, but I share the same feeling when using clifford as a reference to write the `__init__.py` of GAlgebra.
My two cents:
> 1. It's better to split them into a few smaller files but not too many to confuse developers.
> 2. It's better to discuss the logical structure of the split upfront to reach a consensus.
> 3. If it's not a major issue yet, then the split needs to be weighed against better preservation of development history since splitting breaks history like for `git blame`.

Speaking with @hugohadfield, I think we were agreed that this PR represents an improvement over the status quo, so going ahead and merging

PR #153, opened 2019-10-21
Add a linter step that runs flake8

eric-wieser:
Only way to test CI is to make a commit...


eric-wieser:
Seems to work:

image

PR #156, opened 2019-10-22
Fix `Layout.blades_of_grade(0)` to return the right type

eric-wieser:
Previously this returned a single float, but it should have returned a list of multivectors

This also changes `clifford.bases` to include the scalar in its returned dictionary.
This seems eminently reasonable, since the scalar is itself a base.
The only weirdness is that `locals().update(layout.blades())` will now set the variable `''` to something, but nothing can read that anyway.

Adds some type annotations while we're here.

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -39,7 +39,7 @@ Functions
 from functools import reduce
 import os
 import itertools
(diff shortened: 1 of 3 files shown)


eric-wieser:
CI passed fully (!), lets put this in

PR #157, opened 2019-10-22
Turn on flake8 for a bunch more files

eric-wieser:


eric-wieser:
Travis mostly passes, and flake8 definitely does - merging

PR #161, opened 2019-10-22
Avoid type promotion in `MultiVector.__invert__`

eric-wieser:
Also adds better tests for checking for type promotion.


eric-wieser:
In what I can only describe as shocking, the azure builds passed!

PR #162, opened 2019-10-22
Change `isBlade` to work on degenerate metrics

### LINKED ISSUE #145: isBlade fails on degenerate metrics
```python
>>> import clifford
>>> layout, _ = clifford.Cl(3, 0, 1, firstIndex=0)
>>> locals().update(layout.blades)
>>> e0.isBlade()
False
```

As discovered by @enkimute

eric-wieser:
```python
>>> import clifford
>>> layout, _ = clifford.Cl(3, 0, 1, firstIdx=0)
>>> locals().update(layout.blades)
>>> e0.isBlade()
True
>>> e0.isBlade(invertible=True)
False
```

Fixes #145.

---

This exposes that `.factorise()` and `.basis()` both fail on non-invertible blades. I'm not sure exactly how to fix that, other than being sure I don't want to do it in this PR.

### DIFF
--- a/clifford/_multivector.py
+++ b/clifford/_multivector.py
@@ -2,6 +2,7 @@ import numbers
 import math
 from typing import List, Set, Tuple, Union
 import warnings
+import operator
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


clifford/_multivector.py
```diff
@@ -651,33 +652,45 @@ def isScalar(self) -> bool:
...
         if not np.sum(np.abs((Vhat*Vinv).value - (Vinv*Vhat).value)) < 0.0001:
             return False
 
         # applying a versor (and hence an invertible blade) to a vector should
         # not change the grade
         if not all(
-            grades_present(Vhat*e*Vrev, 0.000001) == {1}
+            grade_cmp(grades_present(Vhat*e*Vrev, 0.000001), {1})
             for e in cf.basis_vectors(self.layout).values()
         ):
```
eric-wieser: This condition is relaxed to allow only the removal of grades, but not the addition of any new ones

clifford/_multivector.py
```diff
@@ -651,33 +652,45 @@ def isScalar(self) -> bool:
...
+        elif invertible:
+            return False
+        else:
+            Vinv = Vrev
 
         # Test if the versor inverse (~V)/(V * ~V) is truly the inverse of the
         # multivector V
-        if grades_present(Vhat*Vinv, 0.000001) != {0}:
+        if not grade_cmp(grades_present(Vhat*Vinv, 0.000001), {0}):
             return False
```
eric-wieser: This condition is relaxed to allow `Vhat*Vinv == 0` to be considered

clifford/_multivector.py
```diff
@@ -651,33 +652,45 @@ def isScalar(self) -> bool:
...
         Vrev = ~self
-        Vinv = Vrev/(self*Vrev)[0]
+        mag = (self*Vrev)[()]
+        if mag != 0:
+            # not strictly necessary, just scales to make eps appropriate below
+            Vinv = Vrev / mag
+        elif invertible:
+            return False
+        else:
+            Vinv = Vrev
```
eric-wieser: Changes here are to:

* bail out early without a divide-by-zero if not allowing non-invertible blades
* avoiding the division if it is zero, since that shouldn't affect the results below significantly
* Use `[()]` instead of `[0]`, which is a little more reliable for pulling out the scalar part

PR #165, opened 2019-10-24
Use lru_cache to speed up tests

eric-wieser:
Let's see if this makes them faster


eric-wieser:
This turned out to barely help at all :(

eric-wieser:
Let's kill this patch, using fixtures in #249 did the trick

PR #166, opened 2019-10-24
Apply code and math formatting to docs

eric-wieser:
Will use CI to check this looks right

PR #167, opened 2019-10-24
Use pyganja

eric-wieser:
Builds on top of #166, pushing now to test it works on readthedocs

### DIFF
--- a/docs/Example 1 Interpolating Conformal Objects.ipynb
+++ b/docs/Example 1 Interpolating Conformal Objects.ipynb
@@ -105,22 +105,31 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


hugohadfield:
Looks cool, but the build is broken? https://external-builds.readthedocs.io/html/clifford/167/Example%201%20Interpolating%20Conformal%20Objects.html

eric-wieser:
I think that's an upstream readthedocs bug. It works locally, with the pyganja changes.

eric-wieser:
Going to just merge this, its doc-only, and the ReadTheDocs CI stuff isn't mature enough to work for these filenames. Let's just see if it works.

eric-wieser:
Crap:

    ModuleNotFoundError: No module named 'pyganja'

PR #168, opened 2019-10-25
Regenerate the sphinx makefiles with Sphinx 2.2.0

eric-wieser:
This makes them a lot simpler

Will merge if RTD CI passes


eric-wieser:
Travis CI is uninteresting, and this works on both read-the-docs and locally. Merging...

PR #178, opened 2019-10-28
Add missing change from gh-172

eric-wieser:
Not sure how this got lost

xref #172


eric-wieser:
Got past the install step, let's consider this working

Issue #181, opened 2019-10-28
cga.Dilation(mv) crashes with TypeError:


eric-wieser:
```python
>>> cga = CGA(3)
>>> mv = cga.dilation()
>>> cga.dilation(mv)
TypeError: unorderable types: MultiVector() < int()
```

Issue #180, opened 2019-10-28
cga.Flat.__repr__ crashes with AttributeError


eric-wieser:
```python
    def __repr__(self) -> str:
>       return '%i-Flat' % self.dim
E       AttributeError: 'Flat' object has no attribute 'dim'
```

PR #183, opened 2019-10-28
Cleanup and write tests for clifford.cga

eric-wieser:
This doesn't attempt to fix any bugs, but at least converts unknown bugs into known bugs. The bugs this exposed are:

* #100
* #180
* #181
* #182
* #184
* #185

### DIFF
--- a/clifford/cga.py
+++ b/clifford/cga.py
@@ -76,6 +76,10 @@ class CGAThing(object):
 
     maps versor product to `__call__`.
     '''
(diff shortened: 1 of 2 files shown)


eric-wieser:
Travis passes, coverage increased, and there are open issues for all the xfailing tests. If the tests seem like nonsense, we can fix them in a follow-up / discuss in the open bugs

arsenovic:
nice job on this.

PR #186, opened 2019-10-29
Add a travis run without JIT to improve coverage

eric-wieser:
This will likely be super slow


eric-wieser:
Fails due to https://github.com/numba/numba/issues/4759

PR #188, opened 2019-10-29
Remove bare except

eric-wieser:
Many of these were used for debugging purposes in tests (and mostly meant AssertionError).

These have also be changed to use `raise` to re-raise the exception, which results in a cleaner traceback in pytest.
Note that it's still possible to usefully debug these tests by running pytest with `--pdb`.


eric-wieser:
Seems to work fine, merging...

PR #189, opened 2019-10-29
Add primitive numba support for Layout and MultiVector

eric-wieser:
This implements the minimal operations to allow numba functions like:
```
@numba.njit
def wedge3(a, b, c):
    ret_value = omt_func(a.value, omt_func(b.value, c.value))
    return MultiVector(a.layout, ret_value)
```

This means we can drop all of our `func` / `val_func` duality (in a future commit).

This implementation contains some patterns that do not match the standard numba way of doing things. I've tried to add comments where we diverge.

### DIFF
--- a/clifford/numba/_multivector.py
+++ b/clifford/numba/_multivector.py
@@ -44,10 +44,23 @@ class MultiVectorType(types.Type):
         return LayoutType()
 
 
-@numba.extending.typeof_impl.register(MultiVector)
-def _typeof_MultiVector(val: MultiVector, c) -> MultiVectorType:
-    val._numba_type_ = MultiVectorType(dtype=val.value.dtype)
-    return val._numba_type_
+# The docs say we should use register a function to determine the numba type
+# with `@numba.extending.typeof_impl.register(MultiVector)`, but this is way
+# too slow (https://github.com/numba/numba/issues/5839). Instead, we use the
+# undocumented `_numba_type_` attribute, and use our own cache. In future
+# this may need to be a weak cache, but for now the objects are tiny anyway.
+_cache = {}
+
+@property
+def _numba_type_(self):
+    dt = self.value.dtype
+    try:
+        return _cache[dt]
+    except KeyError:
+        ret = _cache[dt] = MultiVectorType(dtype=dt)
+        return ret
+
+MultiVector._numba_type_ = _numba_type_
 
 
 @numba.extending.register_model(MultiVectorType)
--- a/clifford/numba/_multivector.py
+++ b/clifford/numba/_multivector.py
@@ -121,10 +121,13 @@ def box_MultiVector(typ: MultiVectorType, val: NativeValue, c) -> MultiVector:
     mv = cgutils.create_struct_proxy(typ)(c.context, c.builder, value=val)
     mv_obj = c.box(typ.value_type, mv.value)
     layout_obj = c.box(typ.layout_type, mv.layout)
-    class_obj = c.pyapi.unserialize(c.pyapi.serialize_object(MultiVector))
+
+    # All the examples use `c.pyapi.unserialize(c.pyapi.serialize_object(MultiVector))` here.
+    # Doing so is much slower, as it incurs pickle. This is probably safe.
+    class_obj_ptr = c.context.add_dynamic_addr(c.builder, id(MultiVector), info=MultiVector.__name__)
+    class_obj = c.builder.bitcast(class_obj_ptr, c.pyapi.pyobj)
     res = c.pyapi.call_function_objargs(class_obj, (layout_obj, mv_obj))
     c.pyapi.decref(mv_obj)
-    c.pyapi.decref(class_obj)
     c.pyapi.decref(layout_obj)
     return res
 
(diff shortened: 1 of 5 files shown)


examples/test_numba_extension.py
```diff
@@ -0,0 +1,16 @@
...
+from clifford._numba_multivector import MultiVectorType
+
+from numba.extending import lower_getattr
+from numba import types
+from clifford import MultiVector
+import numpy as np
+
+@numba.njit
+def add(a):
+    return MultiVector(a.layout, a.value)
```
eric-wieser: Something screwy is going on here - if I change this to `a.value * 2`, I get memory filled with `0xDE` . I think the array is being freed or something?

eric-wieser:
Blocked by https://github.com/asodeur/numba-passthru/pull/3

eric-wieser:
`lower_constant` is missing from this patch, and is pretty essential for this to be useful.

clifford/numba/_multivector.py
```diff
@@ -0,0 +1,136 @@
...
+@property
+def _numba_type_(self):
+    dt = self.value.dtype
+    try:
+        return _cache[dt]
+    except KeyError:
+        ret = _cache[dt] = MultiVectorType(dtype=dt)
+        return ret
+
+MultiVector._numba_type_ = _numba_type_
```
eric-wieser, line 63: xref https://github.com/numba/numba/issues/5839

eric-wieser: @hameerabbasi, you might want to use a similar approach in `sparse`, it gives a sizeable performance boost.

clifford/numba/_multivector.py
```diff
@@ -0,0 +1,136 @@
...
+@numba.extending.box(MultiVectorType)
+def box_MultiVector(typ: MultiVectorType, val: NativeValue, c) -> MultiVector:
+    mv = cgutils.create_struct_proxy(typ)(c.context, c.builder, value=val)
+    mv_obj = c.box(typ.value_type, mv.value)
+    layout_obj = c.box(typ.layout_type, mv.layout)
+
+    # All the examples use `c.pyapi.unserialize(c.pyapi.serialize_object(MultiVector))` here.
+    # Doing so is much slower, as it incurs pickle. This is probably safe.
+    class_obj_ptr = c.context.add_dynamic_addr(c.builder, id(MultiVector), info=MultiVector.__name__)
+    class_obj = c.builder.bitcast(class_obj_ptr, c.pyapi.pyobj)
```
eric-wieser, line 128: This is the other non-idiomatic optimization. @asodeur, have you used this in any of your work?

asodeur: Initially used this but later switched to the (de)serialize approach because caching is important in our use-case. The only thing to watch-out for is you have to manage the lifetime of the object (`MultiVector` looks immortal so that should be ok). 

There is also [env_manager](https://github.com/numba/numba/pull/4929) that will manage lifetimes for you. Usage is like:
```python
 obj_ptr = c.env_manager.read_const(c.env_manager.add_const(obj))
```
Again does prevent caching and you cannot call the boxer from within jitted functions (via `objmode` etc).

eric-wieser: Thanks for the reference, that issue seems to cover exactly the cases I care about.

>  you cannot call the boxer from within jitted functions (via objmode etc).

Is this a bug in numba, or is this by design?

clifford/numba/_multivector.py
hameerabbasi: Compilation time, call time outside Numba, or call time within Numba? I assume call time outside Numba.

eric-wieser: The latter, time to enter a numba function that accepts a sparse matrix. Here at Clifford we never actually used the sparse support I wrote, because the performance cost was seemingly too high.

clifford/numba/_multivector.py
asodeur: This seems to be by design. The `env_manager` is tied to the Python wrapper of the compiled function. If you are calling the function from another jitted function it might not be initialized, yet.

I think what is needed would be something like `env_manager` on LLVM module level, ie a list you could append `PyObject*`s to that gets cleaned-up when the LLVM module is no longer needed.

clifford/test/test_numba_extensions.py
```diff
@@ -0,0 +1,50 @@
...
+        def negate(a):
+            return cf.MultiVector(a.layout, -a.value)
+
+        n_e1 = negate(e1)
+        assert n_e1.layout is e1.layout
+        assert n_e1 == -e1
+
+        @numba.njit
+        def add(a, b):
+            return cf.MultiVector(a.layout, a.value + b.value)
```
hugohadfield, line 39: So the plan from here towards having code like:
```
def add(a, b):
    return a + b
```
is to register overloaded function add for the MultiVector type and then use the AST rewrite for the operator overload?

eric-wieser: There's no need for an ast rewrite at all, but we'll need to `@overload(operator.add)`

PR #190, opened 2019-10-30
TST: io: Use temporary files for testing

eric-wieser:
This avoids them being accidentally checked into git

Also remove the * imports and the flake8 config that permits them, for clarity

### DIFF
--- a/clifford/test/test_io.py
+++ b/clifford/test/test_io.py
@@ -1,11 +1,15 @@
-import unittest
-from clifford import *
-from clifford.g3c import *
-from clifford.tools.g3c import *
-from clifford.io import *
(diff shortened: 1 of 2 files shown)


clifford/test/test_io.py
```diff
@@ -1,22 +1,26 @@
-import unittest
-from clifford import *
-from clifford.g3c import *
-from clifford.tools.g3c import *
-from clifford.io import *
+import numpy as np
```
eric-wieser: Surprise: `from clifford import *` imports numpy!

eric-wieser:
Something is up with conda, but it seems to be an upstream issue

PR #191, opened 2019-11-06
Add pyganja visualizations to 2D CGA

eric-wieser:
Note that this is full with calls to `homo` since ganja.js and clifford disagree on n0 and ninf (cc @enkimute)

https://clifford--191.org.readthedocs.build/en/191/ConformalGeometricAlgebra.html

### DIFF
--- a/docs/ConformalGeometricAlgebra.ipynb
+++ b/docs/ConformalGeometricAlgebra.ipynb
@@ -55,13 +55,12 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": false
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "from numpy import pi,e\n",
     "from clifford import Cl, conformalize\n",
+    "from pyganja import draw, GanjaScene\n",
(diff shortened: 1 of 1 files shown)


eric-wieser:
Some conversation [from 2019-11-06 on slack](https://clifforddev.slack.com/archives/CH0NA10LE/p1573046441427000):

@arsenovic
> while i llike the idea of this i think these visualizaitons add nothign and make it more complicated to read the docs

@eric-wieser 
> That's a fair comment

>Would you feel differently if the visualizations were smaller?

@arsenovic

> mmm maybe we could add them at the bottom and have them do something useful?

> i guess it depends on why we are adding visualizations, is it  teach how to use them with conformal objects?

> maybe it would be better if we created a line and then transformed it,  perhaps easier to appreciate than points moving around

@eric-wieser

> I agree, or even had a collection of objects and transformed all of them

---

Now that we have a visualization example (#235), there's less of a need to include that in this notebook. It would still be good to add some transformation visualizations, but I think I agree this is no longer the best way to do it. Since this PR has conflicts. let's just close it.

PR #193, opened 2019-11-07
Replace manual use of `(~B_c*(1/(B_c*~B_c)[0])` with a call to normalInv

eric-wieser:
Since `normalInv` does extra checks, this also introduces a flag to disable those extra checks.

This also speeds up `inv()` - it no longer performs the multiplication `self * ~self` twice.

### DIFF
--- a/clifford/_multivector.py
+++ b/clifford/_multivector.py
@@ -669,7 +669,7 @@ class MultiVector(object):
         :math:`\frac{M}{|M|}` up to a sign
         """
 
-        return self / np.sqrt(abs(self.mag2()))
+        return self / abs(self)
 
     def leftLaInv(self) -> 'MultiVector':
         """Return left-inverse using a computational linear algebra method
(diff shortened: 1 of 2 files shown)


eric-wieser:
Timings before:
```python
In [1]: from clifford.g3c import *

In [2]: a = (e1 + e3 + e123)

In [3]: %timeit a.normalInv()
133 µs ± 5.08 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [4]: %timeit a.inv()
252 µs ± 19.7 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [5]: %timeit ~a*(1/(a*~a)[0])
38.9 µs ± 1.82 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Timings after:
```python
In [1]: from clifford.g3c import *

In [2]: a = (e1 + e3 + e123)

In [3]: %timeit a.normalInv(check=False)
34 µs ± 1.05 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [4]: %timeit a.normalInv()
131 µs ± 5.79 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [5]: %timeit a.inv()
139 µs ± 10.4 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

`.normalInv(check=False)` is actually faster than the full expression, since it doesn't compute `~a` twice!

PR #195, opened 2019-11-08
Add a tool for classifying arbitrary blades

### LINKED ISSUE #192: Improve upon interpret_multivector_as_object
It would be valuable pedagogically to be able to convert arbitrary blades into a description of what they represent geometrically, especially in CGA.

The beginnings of an approach looks like:
<details>

<summary>collapsed, for brevity</summary>

```python
class Object:
    def __init__(self, magnitude=1):
        self.magnitude = magnitude

    def _repr_skip_members(self):
        if self.magnitude == 1:
            return {'magnitude'}
        return {}

    def __repr__(self):
        members = self.__dict__.copy()
        for name in self._repr_skip_members():
            members.pop(name)
        return "{}({})".format(type(self).__name__, ", ".join(
            "{}={!r}".format(k, v) for k, v in members.items()
        ))


class DualPlane(Object):
    def __init__(self, normal, distance, magnitude=1):
        super().__init__(magnitude)
        self.normal = normal
        self.distance = distance

class InfinitePoint(Object): pass

class DualSphere(Object):
    def __init__(self, location, radius, magnitude=1):
        super().__init__(magnitude)
        self.location = location
        self.radius = radius

class Point(DualSphere):
    def __init__(self, location, magnitude=1):
        super().__init__(location, 0, magnitude=magnitude)
    def _repr_skip_members(self):
        return super()._repr_skip_members() | {'radius'}

def classify(x):
    grades = x.grades()
    if grades == {1}:
        # GA4CS: Table 13.2
        if x | einf == 0:
            d = x | -eo
            n = x - d*einf
            mag = n*n
            if mag != 0:
                mag = math.sqrt(mag)
                return DualPlane(n/mag, d/mag, magnitude=mag)
            else:
                return InfinitePoint(magnitude=d)
        else:
            mag = x | -einf
            x = x / mag
            x_ninf = x | -eo
            location = x - eo - x_ninf*einf
            rad2 = location*location - 2*x_ninf
            if rad2 == 0:
                return Point(location, magnitude=mag)
            else:
                return DualSphere(location, math.sqrt(rad2), magnitude=mag)

    return "unknown {}-blade".format(grades.pop())
```

</details>

Working as:
```python
>>> from clifford.g2c import  *
>>> classify(e1)
DualPlane(distance=0, normal=(1.0^e1))
>>> classify(up(e1))
Point(location=(1.0^e1))
>>> classify((up(e1)^up(e2)^up(e1+e2)).dual())
DualSphere(radius=0.7071067811865476, location=(0.5^e1) + (0.5^e2))
>>> classify(einf)
InfinitePoint()
>>> classify(einf*2)
InfinitePoint(magnitude=2.0)
```
This overlaps a fair amount with the `clifford.cga` module, and doesn't support any higher dimensional objects yet.

eric-wieser:
Closes #192.

To resolve:
* [ ] Determine what merging with `clifford.cga` would look like.
* [x] Add tests
* [x] Add sphinx documentation (Done, [here](https://external-builds.readthedocs.io/html/clifford/195/generated/clifford.tools.classify.html#module-clifford.tools.classify))
* [x] Work out how to make this reusable for g2c as well?
  * Needs gh-197

### DIFF
--- a/clifford/tools/classify.py
+++ b/clifford/tools/classify.py
@@ -1,0 +1,472 @@
+"""
+Tools for interpreting conformal blades
+
+.. currentmodule:: clifford.tools.classify
+
+.. autofunction:: classify
+
+The return type is a :class:`Blade`:
+
+.. autoclass:: Blade
+    :members:
+
+The subclasses below are the four categories to which all blades belong, where
+:math:`E` is a euclidean blade, and :math:`T_p[X]` represents a translation of
+the conformal blade :math:`X` by the euclidean vector :math:`p`.
+
+.. autoclass:: Direction
+
+.. autoclass:: Flat
+
+.. autoclass:: DualFlat
+
+.. autoclass:: Round
+
+These can be constructed directly, and will attempt to show a grade-specific
+interpretation::
+
+    >>> from clifford.g3c import *
+    >>> Round(location=e1, direction=e1^e2, radius=1)
+    Circle(location=(1^e1), direction=(1^e12), radius=1)
+    >>> Round(location=e1, direction=e1^e2^e3, radius=1)
+    Sphere(location=(1^e1), direction=(1^e123), radius=1)
+
+
+Aliased types
+-------------
+
+In addition, aliases are created for specific grades of the above types, with
+more convenient names:
+
+.. autoclass:: Tangent
+    :show-inheritance:
+
+.. autoclass:: Point
+    :show-inheritance:
+
+.. autoclass:: PointFlat
+    :show-inheritance:
+
+.. autoclass:: Line
+    :show-inheritance:
+
+.. autoclass:: Plane
+    :show-inheritance:
+
+.. autoclass:: PointPair
+    :show-inheritance:
+
+.. autoclass:: Circle
+    :show-inheritance:
+
+.. autoclass:: Sphere
+    :show-inheritance:
+
+.. autoclass:: InfinitePoint
+    :show-inheritance:
+
+
+
+"""
+import math
+import sys
+from typing import Optional
+import copyreg
+
+from clifford._multivector import MultiVector
+from clifford._conformal_layout import ConformalLayout
+
+__all__ = [
+    'classify',
+
+    'Blade',
+    'Direction',
+    'Flat',
+    'DualFlat',
+    'Round',
+    'Tangent',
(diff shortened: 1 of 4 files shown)


clifford/tools/g3c/classify.py
```diff
@@ -0,0 +1,238 @@
...
+def classify(x) -> Blade:
+    """
+    Classify a multivector into a parameterized geomtric description.
+
+    Derived from GA4CS Table 14.1
+
+    Example usage::
+
+        >>> classify(e1)
+        DualFlat[1](location=0, direction=-(1.0^e23))
```
eric-wieser: This naming seems a little unhelpful to me - is there a better interpretation of these conformal blades?

clifford/tools/classify.py
```diff
@@ -0,0 +1,361 @@
+"""
+Tools for interpreting conformal blades
```
eric-wieser: Not sure whether `clifford.tools.classify` is the right name for this.

clifford/tools/classify.py
```diff
@@ -0,0 +1,412 @@
...
+.. autoclass:: Round
+
+These can be constructed directly, and will attempt to show a grade-specific
+interpretation::
+
+    >>> from clifford.g3c import *
+    >>> Round(location=e1, direction=e1^e2, radius=1)
+    Circle(location=(1^e1), direction=(1^e12), radius=1)
+    >>> Round(location=e1, direction=e1^e2^e3, radius=1)
+    Sphere(location=(1^e1), direction=(1^e123), radius=1)
```
eric-wieser: Updated after some confusion in a demo with Joan

eric-wieser:
Got broad approval from @arsenovic via slack:

> i really like the structure of it and use of inheritance, like PointPair-> Round(2)
>
> i think that is really nice
>
> docs are really clean and nice use of latex
>
> i think we will see how good the structure is when it is used in some examples
>
> the actual class'ing of the code is foreign to me.
>
> [...]
>
> i think its great

(posting here for posterity, in case someone asks "how did anyone think this was a good idea" in a few years)

Travis passes now that I backed out the change to `interpret_multivector_as_object`, which is apparently more resilient to numerical error than this approach.

Issue #34, opened 2018-07-17
domains of conformalize up()/down()  algebra 


arsenovic:
i think the `up` and `down` methods of `conformalize()` should  take/return vectors in the different algebras: 
* `up` should operate on vectors in  ga, and 
* `down` should operate on vectors in cga

we could overload each method  to work on vectors in either algebra. here is a quick way to do it,

    from numpy import zeros
    from clifford import Cl

    ga,blades = Cl(2)
    cga,blades,stuff = conformalize(ga)
    def up(x):
        old_val = x.value
        new_val = zeros(cga.layout.gaDims)
        new_val[:len(old_val)] = old_val
        x_in_cga =  cga.MultiVector(value=new_val)
        #do normal up method on x_in_cga

and similar for `down` .what do you think?   also, we need a consistent name for the algebra that is conformalized, like original, lower, whatever.

moble:
I really like this idea, and [your PR](/pygae/clifford/pull/35) makes it look like an elegant approach, leaving room for helpful specializations and generally more descriptive code.

Now, to spitball some words that could be used to define the different methods and so on: In the literature, you'll frequently see discussion of the progression from Cartesian, to projective, and then conformal *models* of the same underlying Euclidean geometry.  So I wonder if `conformal` and `cartesian` might be more descriptive and/or readily understood than `up` and `down`, respectively — and leave room for `projective` when people want to use that model.  Maybe more generally, if you think of one space as a subspace of another, an element of the subspace is mapped into the superspace by "inclusion" or sometimes even "promotion", and mapped back to the subspace by "projection", "restriction", or "demotion".  Also, when you have some structure over a space, you frequently call the original space the "base" space.

hugohadfield:
I think in general we need to address the problems of mixing multivectors from multiple algebras. The most common form of this I run into is probably the cartesian/conformal pair but it is likely to appear in a variety of places in which we want to use code written for different algebras where one is a sub-algebra of another.
I don't really know what the solution to this is and haven't touched it at all so far for fear opening a big can of worms that I don't have the time to get really stuck into... currently I find working entirely in (4,1) works well for all my (3,0) and (4,1) code and prevents any potential headaches caused by forgetting what layout is associated with what mv.

moble:
[...]

PR #196, opened 2019-11-09
Move random functions to their own file

eric-wieser:
This makes tools/g3c a slightly more manageable length.

It also makes it easier to switch to the new numpy random API in future, which would solve our non-deterministic CI woes.


eric-wieser:
Except I screwed up and CI fails.

Some of the functions in `tools.g3` seem to assume `len(Multivector.value) == 32)` - which actually makes them g3c tools, right?

hugohadfield:
Yeah this brings us to the difficult problem of moving between algebras... The tools in tools.g3 actually just use clifford.g3c as I had no really good way of mixing algebras. This is obviously not a dream situation. @arsenovic has a blade map setup but to my shame I never worked out how to use it

hugohadfield:
Maybe it's time we opened this can of worms again https://github.com/pygae/clifford/issues/34

Issue #37, opened 2018-07-23
make Cl() return layout only , why do we have Cl()?


arsenovic:
since `blades` are in layout, `Cl()` doesnt need to return a tuple of `layout, blades`, rather it could just return layout.  this would make for mildly cleaner code. this will break things, so its worth considering. also i think we could consider changing `Cl()` to `ga()`, so we can write 

`g3 = ga(3)`

could even remove the `a`, so its `g3 =g(3)`, since its a method. which brings up the point, why are we not just using the layout constructor?  couldnt the stuff in `Cl()`  be moved into `Layout.__init__`? and why isnt `Layout` called `Ga` or `G` or whatever,

hugohadfield:
This definitely sounds sensible but isn't really a pain point for me atm, it works at the moment even if it is kinda weirdly done

eric-wieser:
What does `Cl` stand for?

PR #197, opened 2019-11-10
Make `conformalize` return a subclass of `Layout`.

eric-wieser:
This means that `layout.up` and `layout.einf` etc are available, which makes it easier to write code that works for both g3c and g2c.

In order to make this change, most of the interesting logic in `Cl` and `conformalize` moved to private classmethods in `Layout` and `ConformalLayout`.
In future, we could consider making these methods public, and maybe deprecating the public ones.

Relates somewhat to #37

### DIFF
--- a/clifford/_conformal_layout.py
+++ b/clifford/_conformal_layout.py
@@ -1,0 +1,87 @@
+import numpy as np
+
+from ._layout import Layout
+from ._multivector import MultiVector
+
+
+class ConformalLayout(Layout):
+    r"""
+    A layout for a conformal algebra, which adds extra constants and helpers.
+
+    Typically these should be constructed via :func:`clifford.conformalize`.
+
+    Attributes
+    ----------
+    ep : MultiVector
+        The first added basis element, :math:`e_{+}`, usually with :math:`e_{+}^2 = +1`
+    en : MultiVector
+        The first added basis element, :math:`e_{-}`, usually with :math:`e_{-}^2 = -1`
+    eo : MultiVector
+        The null basis vector at the origin, :math:`e_o = 0.5(e_{-} - e_{+})`
+    einf : MultiVector
+        The null vector at infinity, :math:`e_\infty = e_{-} + e_{+}`
+    E0 : MultiVector
+        The minkowski subspace bivector, :math:`e_\infty \wedge e_o`
+    I_base : MultiVector
+        The pseudoscalar of the base ga, in cga layout
+    """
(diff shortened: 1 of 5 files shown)


clifford/_layout.py
```diff
@@ -157,6 +157,18 @@ def __init__(self, sig, bladeTupList, firstIdx=0, names=None):
...
+    def _from_sig(cls, sig=None, *, firstIdx=1, **kwargs):
+        """ Factory method that uses sorted blade tuples.  """
+        bladeTupList = cf.elements(len(sig), firstIdx)
+
+        return cls(sig, bladeTupList, firstIdx=firstIdx, **kwargs)
+
+    @classmethod
+    def _from_Cl(cls, p=0, q=0, r=0, **kwargs):
+        """ Factory method from a ${Cl}_{p,q,r}$ notation """
+        return cls._from_sig([0]*r + [+1]*p + [-1]*q, **kwargs)
```
eric-wieser: For another time - decide if these make better sense as a public API than the current `clifford.Cl`, and if so come up with succinct names (and remove the leading `_`

clifford/_conformal_layout.py
```diff
@@ -0,0 +1,87 @@
...
+        self._base_layout = layout
+
+        basis_vectors = self.basis_vectors
+        added_keys = sorted(basis_vectors.keys())[-2:]
+        ep, en = [basis_vectors[k] for k in added_keys]
+
+        # setup  null basis, and minkowski subspace bivector
+        eo = .5 ^ (en - ep)
+        einf = en + ep
+        E0 = einf ^ eo
```
eric-wieser: Note that this is just copied from the old code - I haven't changed any definitions here

eric-wieser:
@arsenovic: does this seem reasonable, or do you think it overlaps too much with `clifford.cga`?

clifford/_conformal_layout.py
arsenovic: i like this change.

clifford/_layout.py
arsenovic: i like them as public methods. not sure of a good name for the `from_pqr()`, but that it used most i would guess.

eric-wieser: Perhaps just `Layout.Cl(...)`

eric-wieser:
Let's put this in - even if it does overlap with `cga.CGA`, we're already relying on `Layout` to have `einf` and `eo` members elsewhere. We can deal with any necessary merging later.

PR #199, opened 2019-11-11
Fix pow as per Daniil Slavins PR

hugohadfield:


clifford/_multivector.py
```diff
@@ -273,7 +273,8 @@ def __pow__(self, other) -> 'MultiVector':
         other = int(round(other))
 
         if other == 0:
-            return 1
+            unit_out = self._newMV(dtype=self.value.dtype) + 1
+            return unit_out
```
eric-wieser: Shorter as `self.layout.scalar` I think

eric-wieser:
Would be good to add a test

clifford/test/test_clifford.py
```diff
@@ -40,6 +40,14 @@ def test_pseudoscalar(self, algebra):
...
         ps2 = layout.pseudoScalar
         assert np.linalg.matrix_rank(np.column_stack((ps.value, ps2.value))) == 1
+        
+    def test_pow_0(self, algebra):
+        layout, blades = algebra
+        e1 = blades['e1']
+        ret = e1**0
+        assert type(ret) is type(e1)
+        assert ret == 1
+        assert ret.value.dtype == el.value.dtype
```
hugohadfield:
```suggestion
        assert ret.value.dtype == e1.value.dtype
```

eric-wieser:
Fails CI with `clifford/test/test_clifford.py:43:1: W293 blank line contains whitespace`

PR #200, opened 2019-11-13
Remove the repeated member listings from autosummarized modules

eric-wieser:
Since our modules list their members in a structured way within their docstring, we do not need a repeated list of members at the bottom.

Unfortunately, we are not able to specify this template per-autosummary, due to the hack in our conf.py to generate autosummary directives recursively. It seems safe to perform the global override.


eric-wieser:
Readthedocs looks ok, merging...

PR #206, opened 2019-11-19
Fix deployment to only run if tests pass

### LINKED ISSUE #96: Build and release wheel files
We should build and upload wheel files with our releases, would allow us to be hosted on pyiodide
https://alpha.iodide.io/notebooks/2805/

### LINKED ISSUE #179: Travis deploys to pypi as soon as flake8 passes, whether or not the tests actually pass
Somewhat related to #121

eric-wieser:
This also releases a pure-python wheel, just in case that's convenient for some users

This change takes advantage of the yaml features described [here](https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd)

Note that this removes `sudo: true`, which travis ignores anyway\

Closes #179, closes #96

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,52 @@
+# there is nothing special about this key, it's just ignored by travis
+x-clifford-templates:
+  lint_job: &lint_job
+    install:
+      - pip install flake8
+      - python setup.py install
+    script: flake8 clifford
 
-sudo: false
+  test_job: &test_job
+    install:
+      - |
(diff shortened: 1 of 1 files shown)


eric-wieser:
Tested with the test section commented out and no deployment, and vice versa. Let's merge this and attempt a 1.1.2 release with wheels.

eric-wieser:
Attempting a 1.1.2 release here: https://travis-ci.org/pygae/clifford/builds/613976239?utm_medium=notification&utm_source=github_status

PR #209, opened 2019-11-20
Simplification of homo function

hugohadfield:


eric-wieser:
Yep, `x(0).normalInv()` is just a weird way to spell `1 / x[()]`

PR #212, opened 2019-11-29
Add transformation tools

eric-wieser:
These will help with basis changes, most notably handy for CGA.

~~Hopefully the doctests are enough to produce coverage data...~~

With thanks to @enki and @jeremyong for helping me work this out, and @brombo for a very similar class in `galgebra`. Naming suggestions would be appreciated.

### DIFF
--- a/clifford/transformations.py
+++ b/clifford/transformations.py
@@ -1,0 +1,318 @@
+"""
+.. currentmodule:: clifford.transformations
+
+=================================================
+transformations (:mod:`clifford.transformations`)
+=================================================
+
+.. versionadded:: 1.3.0
+
+This module provides some base classes:
+
+.. autodata:: Transformation
+
+.. autoclass:: Linear
+
+.. autoclass:: FixedLayout
+
+And some matrix-backed implementations:
+
+.. autoclass:: LinearMatrix
+
+.. autoclass:: OutermorphismMatrix
+
+It also provides a helper function for the most common use case, converting
+between basis vectors of similar algebras:
+
+.. autofunction:: between_basis_vectors
+
+This module may in future become the home for optimized rotor transformations,
+or non-linear transformations
+"""
+from typing import Dict, Any, Callable
+from abc import abstractmethod
+
+import numpy as np
+
+from ._bit_helpers import set_bit_indices
+from ._layout_helpers import BasisBladeOrder
+from ._multivector import MultiVector
+from ._layout import Layout
+
+
+# keep this in sync with the docstring!
+__all__ = [
+    'Transformation',
+    'Linear',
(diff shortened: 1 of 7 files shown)


clifford/_transformation.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # fill the scalar
+        self[(),()] = 1
+
+        # fill the vectors
+        for i_from in range(layout_from.dims):
+            vector_from = (i_from+layout_from.firstIdx,)
+            for i_to in range(layout_to.dims):
+                vector_to = (i_to+layout_to.firstIdx,)
+                self[vector_to, vector_from] = matrix[i_to, i_from]
```
eric-wieser: Assigning by tuple like this is probably super slow, but optimizing it is a task for later - preferably once we can guarantee the blade order in the `Layout` class.

eric-wieser:
Will come back to this shortly and add `LinearTransformation.from_function(lambda x: ...)`, which can be used as something like:
```
def rotor_to_matrix(R):
    return LinearTransformation.from_function(lambda x: R * x * ~R).vector_matrix()
```

clifford/_transformation.py
eric-wieser: The changes above have been merged since I made this PR - the latest revision makes use of the new features.

hugohadfield:
This looks nice, should hopefully allow us to make some nice tools for swapping between (sub)algebras easily and efficiently as well as swallowing up things like:
https://github.com/pygae/clifford/blob/93b0188652e96c7768a658abea4ffccbb8fc358b/clifford/__init__.py#L107 
 and other assorted linear algebra gubbins in the tools section. 

My only slight concern with it is the name, transformations for me in GA screams rotors and bivectors rather than linear algebra. What would you think about clifford.linalg ?

clifford/transformations.py
```diff
@@ -0,0 +1,312 @@
...
+    def adjoint(self) -> "LinearMatrix":
+        """ The adjoint transformation """
+        return LinearMatrix(self._matrix.T, self.layout_dst, self.layout_src)
+
+
+class OutermorphismMatrix(LinearMatrix):
+    r"""
+    A generalization of a linear transformation to vectors via the outer product.
+
+    Namely, given a linear transformation :math:`F(u) \to v`, this generalize
```
hugohadfield, line 222: this generalize to

hugohadfield: should be 'generalizes'

eric-wieser:
Finally got the link to the apollonius tutorial to work, will merge once CI is done.

eric-wieser:
CI failure is the numba RC issue again, so out of scope.

PR #215, opened 2019-12-09
Make scalar integral rather than floating-point

eric-wieser:
This makes it consistent with layout.pseudoScalar


eric-wieser:
CI failure is a conda / pytest issue, will fix that separately

eric-wieser:
```
ValueError: eig() argument must not cause a domain change
```
Interesting transient error to debug at some point.

Issue #216, opened 2019-12-10
right_complement might be the left_complement


eric-wieser:
See https://github.com/pygae/clifford/pull/217/files#r356344942

PR #217, opened 2019-12-10
Add a test of right_complement

eric-wieser:
Will merge once CI passes - even if this test is wrong, it matches the current behavior so we can use it to detect behavior changes. gh-216 tracks fixing the possible reversion here.

### DIFF
--- a/clifford/test/test_degenerate.py
+++ b/clifford/test/test_degenerate.py
@@ -5,14 +5,13 @@ import math
 import numpy as np
 
 
-class PGATests(unittest.TestCase):
+class TestPGA:
+    layout, _ = Cl(3, 0, 1, firstIdx=0)
 
-    def setUp(self):
-        self.algebras = [Cl(3, 0, 1, firstIdx=0), Cl(4, 0, 1, firstIdx=0)]
-
-    def test_integration(self):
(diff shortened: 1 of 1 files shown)


clifford/test/test_degenerate.py
```diff
@@ -5,14 +5,18 @@
...
-    def test_integration(self):
-        layout = self.algebras[0][0]
-        blades = self.algebras[0][1]
+class TestPGA:
+    layout, _ = Cl(3, 0, 1, firstIdx=0)
+
+    def test_right_complement(self):
+        # TODO[gh-216]: Does this mean we implemented the left complement?
+        for b in self.layout.blades_list:
+            assert b.right_complement() * b == self.layout.pseudoScalar
```
eric-wieser: @enkimute: should this be (the passing):
```suggestion
            assert b.right_complement() ^ b == self.layout.pseudoScalar
```
or even the failing
```suggestion
            assert b ^ b.right_complement() == self.layout.pseudoScalar
```
?

enkimute:
```
assert b ^ b.right_complement() == self.layout.pseudoScalar
assert b.left_complement() ^ b == self.layout.pseudoScalar
```

PR #218, opened 2019-12-10
Move some private functions to `_layout.py`

eric-wieser:
These are used as part of layout construction.

No one should ever be calling these directly, given they are used just to populate `Layout` members


eric-wieser:
CI failure is stochastic, as usual

PR #219, opened 2019-12-11
Correct the sign of the vee product 

eric-wieser:
This uses the complement functions instead of the dual, which is faster.

Additionally, it uses an equivalent of "undual" as the final step, which makes the sign correct

### DIFF
--- a/clifford/_multivector.py
+++ b/clifford/_multivector.py
@@ -237,6 +237,9 @@ class MultiVector(object):
     def right_complement(self) -> 'MultiVector':
         return self.layout.MultiVector(value=self.layout.right_complement_func(self.value))
 
+    def left_complement(self) -> 'MultiVector':
+        return self.layout.MultiVector(value=self.layout.left_complement_func(self.value))
+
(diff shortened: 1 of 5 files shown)


clifford/_multivector.py
```diff
@@ -97,15 +97,26 @@ def exp(self) -> 'MultiVector':
...
+        functions instead, as these work in degenerate metrics like PGA too,
+        and are equivalent but faster in other metrics.
         """
         return self.layout.MultiVector(value=self.layout.vee_func(self.value, other.value))
 
     def __and__(self, other) -> 'MultiVector':
-        """
-        The vee product aka. the meet... To be optimised still
-        """
+        """ Alias for :meth:`~MultiVector.vee` """
```
hugohadfield: meth?

eric-wieser: Sphinx link to a **meth**od

hugohadfield: ah ok, that makes sense

eric-wieser:
Let's see if I can get CI passing...

PR #220, opened 2019-12-11
Fix issues with old pytest version in conda CI

eric-wieser:
This was already fixed for azure in 364c09c8af4eb1b3755ca4cbcc325c66dbd2bcd2, but seems to have come back for travis.


eric-wieser:
Seems to have solved the issue, the conda runs are no longer failing with errors about `tmppath`

PR #224, opened 2020-01-16
WIP: Add a matplotlib renderer for CGA2D and CGA3D 

eric-wieser:
This will be more suited to animation than pyganja.

image

We might decide we don't want to distribute this in clifford, in which case we can move it to its own package.

Still needs a renderer for 3d lines and planes.


eric-wieser:
Closing, let's put this somewhere else (https://github.com/pygae/mpl_toolkits.clifford)

PR #232, opened 2020-02-23
Fix dependency installation on travis

eric-wieser:
It seems that running `setup.py install` without going through `pip` results in the `python_requires` of dependencies being ignored.

Example of a failing job: https://travis-ci.org/pygae/clifford/builds/653883681?utm_source=github_status&utm_medium=notification


eric-wieser:
This fixes the travis builds back from "always broken" to "usually broken"

PR #233, opened 2020-02-23
Move gp, op, and ip to a new `clifford.operator` module

eric-wieser:
The primary goal here is to make them easier to document (they previously were not documented at all).
The names are still perfectly usable from the old location.

Also, fix the `ip` to be the inner product and not the outer product.

image

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -56,6 +56,9 @@ from clifford.io import write_ga_file, read_ga_file  # noqa: F401
 
 from ._version import __version__  # noqa: F401
 
+# For backwards-compatibility. New code should import directly from `clifford.operator`
+from .operator import gp, op, ip  # noqa: F401
+
 _eps = 1e-12            # float epsilon for float comparisons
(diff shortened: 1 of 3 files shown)


clifford/__init__.py
```diff
@@ -850,57 +853,6 @@ def print_precision(newVal):
...
-    """
-    Inner product function
-
-    ip(M, N) =  M | N
-
-    M and N must be from the same layout
-
-    """
-
-    return M ^ N
```
eric-wieser: This line was very wrong!

eric-wieser:
Nothing I can do about the LGTM warning.

Change should be harmless, and CI passes as much as it ever does - will merge without review.
I don't think `clifford.operator` as a parallel to `operator` is a contentious name, and the functions already existed anyway.

clifford/__init__.py
arsenovic: wow what an error!

arsenovic:
i like this change.

PR #235, opened 2020-02-23
Add visualization documentation

eric-wieser:
Will appear at https://clifford--235.org.readthedocs.build/en/235/tutorials/cga/visualization-tools.html


eric-wieser:
Hugo says

> Just reading your new visualisation docs look good

Given the actual ga stuff is easy to change, but the build system is a nightmare, I think I'm going to put this in without further review. Patches to improve the explanations or example objects are more than welcome!

The navigation looks a little janky, but that's hard to fix without https://github.com/sphinx-doc/sphinx/pull/3622:

image

PR #238, opened 2020-02-24
Modify the license to be recognizeable by github

eric-wieser:
Hopefully this should do the trick.

This removes a (probably unintentional) reference to `scikit-rf` from the license, which I assume was copied from another repo.

Works on my fork:

image


arsenovic:
yes this is fine with me.  obviously an error .

eric-wieser:
Works as intended: https://github.com/pygae/clifford/blob/master/LICENSE.txt

PR #241, opened 2020-02-24
Add the start of a tutorial on robot manipulators

eric-wieser:
The chance of this actually rendering properly is approximately 0, for now...

https://clifford--241.org.readthedocs.build/en/241/tutorials/cga/robotic-manipulators.html


eric-wieser:
Will merge as is for now, in case I don't get a change before Game 2020.

If I do, will make revisions in a follow-up

PR #242, opened 2020-02-25
Use overload instead of generated_jit so that we can run with jitting disabled

eric-wieser:
Let's see if CI works. This should dramatically improve coverage if it does.


eric-wieser:
This didn't work because it led to no jitting being done at all!

Closing in favor of gh-256

PR #244, opened 2020-02-26
Add links to docs at the top of every notebook

eric-wieser:
Doing this means that people downloading individual notebooks can get back to the online docs.

These cells will not appear in the web docs.

Copied from the nbsphinx docs, which use the same tricks.


eric-wieser:
Proof that it doesn't show up: https://clifford--243.org.readthedocs.build/en/243/tutorials/g3-algebra-of-space.html

eric-wieser:
Proof that it doesn't show up: https://clifford--243.org.readthedocs.build/en/243/tutorials/g3-algebra-of-space.html

PR #245, opened 2020-02-26
Add links to binder for notebook pages.

eric-wieser:
This config is copied from the nbsphinx project.


eric-wieser:
Docs builds don't seem to be running any more - lets just push this and hope it works.

PR #248, opened 2020-02-26
Move around requirements.txt files to satisfy binder.

eric-wieser:
It seems that binder is pinned on a bad numba version.
Additionally, it is missing the pyganja and mpl_toolkits.clifford dependency.

Two things to check here;

* Docs still can import `pyganja`
* Binder works (try http://mybinder.org/v2/gh/eric-wieser/clifford/fix-requirements.txt?filepath=docs/tutorials/cga/visualization-tools.ipynb)

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,11 +1,10 @@
 # extra requiremenets needed by the docs themselves
 # numpydoc==0.4
-ipython
-matplotlib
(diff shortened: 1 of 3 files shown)


eric-wieser:
Resolved, looks like we don't support `numba==0.45.1` any more. Will merge, since binder now works.

PR #249, opened 2020-02-27
Speed up test_clifford.py by reusing algebras, fix layout.multi vector

eric-wieser:
Locally this took the test from 2m11 to 1m04 for this single test file


eric-wieser:
CI failure irrelevant, approved over the shoulder by Hugo

PR #250, opened 2020-02-27
Convert test_speed to a benchmark.

eric-wieser:
Also, run benchmark tests in a parallel CI run, to save time.


eric-wieser:
CI seems to be passing enough for now...

PR #251, opened 2020-02-28
Make some cleanups to layout helper functions

eric-wieser:
See commits separately.


eric-wieser:
Failures are the usual.

PR #252, opened 2020-02-29
Move more functions from __init__ to _layout

eric-wieser:


hugohadfield:
This looks good, we should check at some point if the parallelisation in the generate tables is actually running..

eric-wieser:
Coverage is low because this code is all jit'd. Going to put this in, it matches the intent of a bunch of previous PRs.

eric-wieser:
> we should check at some point if the parallelisation in the generate tables is actually running..

Agreed - and perhaps whether it makes a difference.

PR #253, opened 2020-03-02
Don't allow nonsense like `v[e1 + e12]`

eric-wieser:
Previously this would only return the lexicographically first basis element if multiple bases were passed in.


eric-wieser:
Seems to work

PR #254, opened 2020-03-02
Remove `del mv[(1,)]`

eric-wieser:
This isn't what the `del` operator is for, and is more clear as `mv[(1,)] = 0` anyway.
`del` is used to remove an entry from a sequence -but this is neither a sequence, nor is setting to 0 the same as removing.

None of our docs or examples appear to use this.

---

If we're worried about this breaking someone, we could deprecate it instead - but I'd be surprised if anyone uses this.

### DIFF
--- a/clifford/_multivector.py
+++ b/clifford/_multivector.py
@@ -418,21 +418,6 @@ class MultiVector(object):
         else:
             self.value[key] = value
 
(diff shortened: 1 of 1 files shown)


eric-wieser:
Let's put this in, @hugohadfield is on board. I'll make sure to include this in the release notes.

PR #255, opened 2020-03-03
Use the outer multiplication table directly in the complement functions

eric-wieser:
This results in faster startup times, as we no longer force the multiplcation function to be generated at startup.


eric-wieser:
[Before](https://travis-ci.org/pygae/clifford/jobs/657674107#L349)

[After](https://travis-ci.org/pygae/clifford/jobs/657821128#L352)

PR #256, opened 2020-03-03
Fix generated_jit to work even if jitting is disabled

eric-wieser:


eric-wieser:
18% increase in coverage in `_layout.py`!

clifford/_utils.py
```diff
@@ -0,0 +1,43 @@
...
+        return (
+            self._rebuild,
+            numba.serialize._reduce_function(self.__func, globs)
+        )
+
+    def __call__(self, *args, **kwargs):
+        return self.__func(*args, **kwargs)
+
+    def __repr__(self):
+        return "_pickleable_function({!r})".format(self.__func)
```
eric-wieser: Tried upstreaming this in https://github.com/numba/numba/pull/5340

@stuartarchibald, do you think something like this would be an acceptable solution to https://github.com/numba/numba/issues/4759?

eric-wieser:
CI failures are the usual

PR #258, opened 2020-03-05
Add tests for ipython pretty-printing

eric-wieser:
This feature was added a little while ago, but without tests.
Also tweaks the printing behavior slightly to more closely match the `repr`, using `str(dtype)` instead of `repr(dtype)`.


eric-wieser:
Another reasonable coverage bump for the core files

Issue #259, opened 2020-03-05
meet is not symmetric


eric-wieser:
I think this is a bug, but I suppose it's possible it's deliberate.

```python
In [30]: from clifford.g3 import *

In [31]: a, b, c = (e1 + e2), (e1 + e3), (e2 + e3)

In [32]: (a^b^c).meet(b^c)
Out[32]: 0  # ???

In [33]: (b^c).meet(a^b^c)
Out[33]: -(2.0^e12) - (2.0^e13) + (2.0^e23)
```

The `meet` operator has been there since the start of time (note that in master `&` is now spelt `*`, and `*` is now spelt `|`):

https://github.com/pygae/clifford/blob/477764f99f79a808fbeac722271f9bc18283fe27/clifford.py#L1515-L1536

@rkern, I notice that your name is on the repo at this point in time do you happen to have any history or memory regarding this line of the release notes?

https://github.com/pygae/clifford/blob/477764f99f79a808fbeac722271f9bc18283fe27/clifford.py#L160

rkern:
That comment referred to the explicit `.inv()` in the calculation, IIRC. I do not remember, however, what references I was working from when I implemented the meet and join and have not worked with GA in a long time, so I don't have much to add here.

PR #261, opened 2020-03-05
Change the meet operator to match ga4cs

eric-wieser:
This likely fixes gh-259.


eric-wieser:
Tests pass, docs are clearer - let's put this in.

PR #263, opened 2020-03-05
Pull more code out of __init__ and into separate files

eric-wieser:
This is motivated partially by getting more meaningful code-coverage metrics, separating things we care about from things that are barely used.
It also hopefully makes resolving circular imports through `__init__.py` a little easier to reason about.

None of this code was modified, it's a straight move followed by the addition of relevant import statements.


eric-wieser:
LGTM is a false alarm, see https://github.com/Semmle/ql/issues/2771

PR #264, opened 2020-03-05
Deprecate clifford.grades_present

eric-wieser:
We already expose this as a method, there's no need to provide two ways to do the same thing.
Note that the method previously did not allow the epsilon to be tweaked.
`eps` was chosen instead of `threshold` as the name to match other existing functions.

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -64,6 +64,7 @@ Miscellaneous functions
 # Standard library imports.
 import os
 import itertools
+import warnings
 from typing import List, Tuple, Set, Container, Dict, Optional
 
 # Major library imports.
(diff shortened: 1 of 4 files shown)


clifford/__init__.py
```diff
@@ -262,15 +262,12 @@ def grade_obj(objin, threshold=0.0000001):
...
-    nonzero = abs(objin.value) > threshold
-    return {
-        grade_i
-        for grade_i, nonzero_i in zip(objin.layout.gradeList, nonzero)
-        if nonzero_i
-    }
+    # for backwards compatibility
+    warnings.warn(
+        "`clifford.grades_present(x)` is deprecated, use `x.grades()` instead. ",
+        "Note that the method uses `clifford.eps()` as the default tolerance."
```
eric-wieser:
```suggestion
        "`clifford.grades_present(x)` is deprecated, use `x.grades()` instead. "
        "Note that the method uses `clifford.eps()` as the default tolerance.",
```

eric-wieser:
Hugo seemed fine with this

PR #265, opened 2020-03-05
Move settings to their own file

eric-wieser:


eric-wieser:
More in the theme of #263, #147

PR #272, opened 2020-03-09
Remove implied __ne__ methods

eric-wieser:
In Python 3, `__ne__(a, b)` is defined by default as `not (a == b)`.
There is no need for us to manually implement the same thing.


eric-wieser:
Coverage drop is expected, removing code that isn't needed but is still tested causes this.

Issue #54, opened 2018-09-24
Finish algebra generation optimisation


hugohadfield:
There is still quite a bit of low hanging fruit performance-wise around the algebra generation setup. Currently we swap between tuples of grades, bitmaps and indices into the mv value array pretty liberally, eliminating this and jitting the remaining functions would allow us to speed up the generation quite a lot as well as allowing us to lean on the automatic parallelisation capabilities in numba and potentially even adding a cuda generation option.

Additionally @enkimute has suggested a nice loop free bitcount method that could add additional performance:
https://github.com/enkimute/ganja.js/issues/17#issuecomment-423962512

eric-wieser:
> Currently we swap between tuples of grades, bitmaps and indices into the mv value array pretty liberally, eliminating this and ...

This is mostly done in #273

eric-wieser:
Instead of using @enkimute's optimization, we now just use `__builtin_popcnt` which might even turn into a single assembly instruction.

PR #273, opened 2020-03-09
Encapsulate ordering and naming conventions in separate classes

eric-wieser:
The end goal here is to be able to more clearly separate which bits of `Layout` are specific to the algebra, and which are specific to the choice of blade names and numbers.

This makes the observation that our blade tuples, eg `(1, 2, 3)` &rarr; `e123` need not be integers at all. We already provide some flexibility via `firstIdx`, but the existence of this field resulted in some weird code that was cleaned up in #277. With this change, it's possible (but perhaps not adviseable) to have `('x', 'y', 'z')` &rarr; `exyz`:
```python
In [1]: from clifford import Layout

In [2]: from clifford._layout import BasisBladeOrder, BasisVectorIds

In [3]: l = Layout([1, 1, 1], ids=BasisVectorIds(['x', 'y', 'z']), order=BasisBladeOrder.shortlex(3))

In [4]: locals().update(l.blades)

In [5]: p = ex ^ ey; p
Out[5]: (1^exy)

In [6]: p['x', 'y']
Out[6]: 1

In [7]: l
Out[7]:
Layout([1, 1, 1],
       ids=BasisVectorIds(['x', 'y', 'z']),
       order=BasisBladeOrder.shortlex(3),
       names=['', 'ex', 'ey', 'ez', 'exy', 'exz', 'eyz', 'exyz'])
```

Perhaps more usefully, it's now legal to write just
```python
In [1]: from clifford import Layout

In [2]: Layout([1, 1, 1])
Out[2]:
Layout([1, 1, 1],
       ids=BasisVectorIds(['x', 'y', 'z']),
       order=BasisBladeOrder.shortlex(3),
       names=['', 'e1', 'e2', 'e3', 'e12', 'e13', 'e23', 'e123'])
```

This partially addresses #54.

The size of this patch is inflated a lot by docs and tests - the actual change is perhaps a net +300 line delta.

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -28,6 +28,17 @@ Whether you construct your algebras from scratch, or use the predefined ones, yo
     Layout
     ConformalLayout
 
+Advanced algebra configuration
+------------------------------
+It is unlikely you will need these features, but they remain as a better
+spelling for features which have always been in ``clifford``.
+
+.. autosummary::
+    :toctree: generated/
+
+    BasisBladeOrder
+    BasisVectorIds
+
 Global configuration functions
 ==============================
 These functions are used to change the global behavior of ``clifford``.
@@ -304,6 +315,7 @@ def val_get_right_gmt_matrix(mt: sparse.COO, x):
 from ._layout import Layout  # noqa: E402
 from ._multivector import MultiVector  # noqa: E402
 from ._conformal_layout import ConformalLayout  # noqa: E402
+from ._layout_helpers import BasisVectorIds, BasisBladeOrder  # noqa: F401
 from ._mvarray import MVArray, array  # noqa: F401
 from ._frame import Frame  # noqa: F401
 from ._blademap import BladeMap  # noqa: F401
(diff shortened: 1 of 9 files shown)


eric-wieser:
Failure is real, turned out I made the following changes to the blade ordering:
```
 (1.0^e5)
 (1.0^e12)
 (1.0^e13)
+(1.0^e14)
+(1.0^e15)
 (1.0^e23)
-(1.0^e14)
 (1.0^e24)
+(1.0^e25)
 (1.0^e34)
-(1.0^e15)
-(1.0^e25)
 (1.0^e35)
 (1.0^e45)
 (1.0^e123)
 (1.0^e124)
+(1.0^e125)
 (1.0^e134)
+(1.0^e135)
+(1.0^e145)
 (1.0^e234)
-(1.0^e125)
-(1.0^e135)
 (1.0^e235)
-(1.0^e145)
 (1.0^e245)
 (1.0^e345)
 (1.0^e1234)
```

eric-wieser:
Performance is pretty untouched:
After:
```
In [3]: from clifford import Cl

In [4]: %timeit Cl(8)
142 ms ± 14.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [5]: %timeit Cl(3)
8.75 ms ± 823 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
Before:
```
In [1]: from clifford import Cl

In [2]: %timeit Cl(8)
139 ms ± 8.03 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: %timeit Cl(3)
9.26 ms ± 1.86 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

eric-wieser:
Let's put this in, since @arsenovic didn't seem opposed over slack.

PR #281, opened 2020-03-12
Tidy up MultiVector construction in clifford.tools

eric-wieser:
See individual commit messages.


eric-wieser:
Coverage is pretty low on these files, but the change was only two find and replaces, so hopefully each one has been tested somewhere.

PR #287, opened 2020-03-23
Make count_set_bits use the intrinsic

eric-wieser:
This doesn't actually seem to effect the timings much at the top level, but this version of `count_set_bits` seemed 25x faster in a contrived benchmark,

Closes gh-54

### DIFF
--- a/clifford/_layout_helpers.py
+++ b/clifford/_layout_helpers.py
@@ -14,22 +14,41 @@ import functools
 import operator
 
 from . import _numba_utils
+import numba
+import numba.extending
+import numba.types
+import numba.config
 
 
-@_numba_utils.njit
-def count_set_bits(bitmap):
-    """
-    Counts the number of bits set to 1 in bitmap
-    """
-    bmp = bitmap
-    count = 0
-    n = 1
-    while bmp > 0:
-        if bmp & 1:
-            count += 1
-        bmp = bmp >> 1
-        n = n + 1
-    return count
+@numba.extending.intrinsic
+def __builtin_popcnt(tyctx, x):
+    """ Emulate clang and GCC's `__builtin_popcnt` """
+    if isinstance(x, numba.types.Integer):
+        def impl(cgctx, builder, sig, args):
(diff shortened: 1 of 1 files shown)


eric-wieser:
Travis didn't start...

eric-wieser:
Timings before: https://travis-ci.org/github/pygae/clifford/jobs/665837589#L406

Timings after: https://travis-ci.org/github/pygae/clifford/jobs/665888351#L364

Barely any difference, likely caused by other factors

eric-wieser:
```
In [9]: @numba.njit
   ...: def test_new(a):
   ...:     for i in range(len(a)):
   ...:         a[i] = __builtin_popcnt(a[i])
   ...:

In [11]: @numba.njit
    ...: def count_set_bits(bitmap):
    ...:     """
    ...:     Counts the number of bits set to 1 in bitmap
    ...:     """
    ...:     bmp = bitmap
    ...:     count = 0
    ...:     n = 1
    ...:     while bmp > 0:
    ...:         if bmp & 1:
    ...:             count += 1
    ...:         bmp = bmp >> 1
    ...:         n = n + 1
    ...:     return count
    ...:

In [12]: @numba.njit
    ...: def test_old(a):
    ...:     for i in range(len(a)):
    ...:         a[i] = count_set_bits(a[i])
    ...:

In [13]: a = np.arange(10000)

# compilation included here
In [14]: %timeit test_old(a)
21.1 µs ± 10.7 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [15]: %timeit test_old(a)
16.2 µs ± 2.03 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)

# compilation included here
In [16]: %timeit test_new(a)
4.59 µs ± 1.41 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [17]: %timeit test_new(a)
3.63 µs ± 597 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

PR #288, opened 2020-03-23
Create .zenodo.json

hugohadfield:
Allows Zenodo to display this library for citation properly


.zenodo.json
```diff
@@ -0,0 +1,34 @@
...
+      "name": "Hadfield, Hugo",
+      "orcid": "0000-0003-4318-050X"
+    },
+    {
+      "affiliation": "University of Cambridge",
+      "name": "Wieser, Eric",
+      "orcid": "0000-0003-0412-4978"
+    },
+    {
+      "affiliation": "Enthough",
```
eric-wieser:
```suggestion
      "affiliation": "Enthought",
```

.zenodo.json
```diff
@@ -0,0 +1,34 @@
...
+      "name": "Kern, Robert"
+    },
+    {
+      "name": "The Pygae Team, "
+    },
+  ],
+  "keywords": [
+  	"python",
+    "clifford algebra",
```
eric-wieser: Mixed tabs and spaces

.zenodo.json
```diff
@@ -0,0 +1,34 @@
...
+      "orcid": "0000-0003-4318-050X"
+    },
+    {
+      "affiliation": "University of Cambridge",
+      "name": "Wieser, Eric",
+      "orcid": "0000-0003-0412-4978"
+    },
+    {
+      "affiliation": "Enthough",
+      "name": "Kern, Robert"
```
eric-wieser: @rkern, let us know if you want an orcid in here, we can add it in a future patch.

.zenodo.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    {
+      "name": "The Pygae Team, "
+    },
```
eric-wieser:
```suggestion
    }
```
Syntax error

.zenodo.json
```diff
@@ -0,0 +1,34 @@
...
+      "affiliation": "University of Cambridge",
+      "name": "Wieser, Eric",
+      "orcid": "0000-0003-0412-4978"
+    },
+    {
+      "affiliation": "Enthough",
+      "name": "Kern, Robert"
+    },
+    {
+      "name": "The Pygae Team, "
```
eric-wieser:
```suggestion
      "name": "The Pygae Team"
```
Sure the comma is a good idea?

hugohadfield: This was predicting bibtex problems with spaces in words... attempted to set it all as a second name, although I might be wrong and it might just work!

.zenodo.json
```diff
@@ -0,0 +1,34 @@
+{
+  "creators": [
```
eric-wieser:
```suggestion
    "license": {
      "id": "BSD-3-Clause"
    }, 
  "creators": [
```

optimistically...

.zenodo.json
eric-wieser:
```suggestion
    "python",
    "clifford algebra",
```

.zenodo.json
```diff
@@ -0,0 +1,34 @@
...
+      "name": "The Pygae Team, "
+    }
+  ],
+  "keywords": [
+  	"python",
+    "clifford algebra",
+    "geometric algebra",
+    "geometry"
+  ],
+  "license": "BSD 3-Clause License",
```
eric-wieser: Hmm... Any idea whether this is correct vs my one above?

.zenodo.json
```diff
@@ -0,0 +1,34 @@
...
+      "orcid": "0000-0002-8599-5873"
+    },
+    {
+      "affiliation": "University of Cambridge",
+      "name": "Hadfield, Hugo",
+      "orcid": "0000-0003-4318-050X"
+    },
+    {
+      "affiliation": "University of Cambridge",
+      "name": "Wieser, Eric",
```
eric-wieser: Note that we can write these as
```suggestion
      "name": "Eric Wieser",
```
and bibtex will behave exactly as before, but zenodo will show our names without a comma.

.zenodo.json
eric-wieser:
```suggestion
      "name": "{The Pygae Team}"
```
might work for zenodo, if it doesn't attempt to escape it.

eric-wieser:
For reference, here's the json schema for zenodo: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/jsonschemas/records/base-v1.0.0.json

PR #292, opened 2020-03-25
Split `MultiVector.__init__` into three static methods

eric-wieser:


eric-wieser:
Upstream issue in sphinx breaks the docs here: patch in sphinx-doc/sphinx#7384

PR #293, opened 2020-03-26
Manually tailor the class documentation pages

eric-wieser:
Notably, this:

* Puts all the member docs in a single page, since there aren't that many of them
* Fixes `Frame` to not show numpy documentation
* Add `MVArray` to the docs
* Makes the documentation for the operator overloads visible.


docs/api/multivector.rst
```diff
@@ -0,0 +1,11 @@
+clifford.MultiVector
+====================
+
+.. currentmodule:: clifford
+
+.. autoclass:: MultiVector
+  :no-inherited-members:
+  :members:
+  :member-order: bysource
```
eric-wieser: This isn't working and I don't know why...

PR #296, opened 2020-03-30
Fix formatting errors in MVArray docs

eric-wieser:


eric-wieser:
Nothing to test here I guess since #293 isn't in yet

PR #297, opened 2020-03-31
Split up table computation to compute each table separately

eric-wieser:
This means that subclasses can override the `gmt` attribute, and will automatically get correct `imt`, `omt`, and `lcmt` tables.

This probably has a small performance penalty, but it should be a constant overhead.


eric-wieser:
CI failure is due to numpy 0.49rc1

PR #299, opened 2020-04-03
Move bit-twiddling helpers to their own file

eric-wieser:
This also deduplicates a repeated bit of code that iterates over bits


eric-wieser:
CI failure is caused by numba again, putting this in as is.

PR #301, opened 2020-04-07
Add LinearMatrix.{from_function, from_rotor}

eric-wieser:
Inspired by conversation on slack, this seems like a fairly common use case, or at least one that is pedagogically useful.

This is tested by its doctests, hopefully. Coverage report will tell.


eric-wieser:
Will fix CI shortly

PR #306, opened 2020-04-08
Add `LinearMatrix.from_mapping`, and replace `BladeMap` with it

eric-wieser:
For the example application in spacetime algebra, `bm(x)` is at least 10x faster.

In theory we could now also write `split` without using `bm` at all:
```python
split = LinearMatrix.from_mapping([
    # even subalgebra of D
    (D.scalar, P.scalar),
    (d01, p1),
    (d02, p2),
    (d03, p3),
    (d12, p12),
    (d23, p23),
    (d13, p13),
    (d0123, p123),

    # split(odd) = split(odd * d0)
    (d0, P.scalar),
    (d1, -p1),
    (d2, -p2),
    (d3, -p3),
    (d012, p12),
    (d013, p13),
    (d023, p23),
    (d123, -p123),
])
```
but I don't have a good enough understanding of what `split.adjoint(p1)` should be to feel comfortable doing this.

### DIFF
--- a/clifford/_blademap.py
+++ b/clifford/_blademap.py
@@ -1,6 +1,11 @@
-class BladeMap(object):
+from .transformations import LinearMatrix
+
+class BladeMap:
     '''
-    A Map Relating Blades in two different algebras
+    A Map Relating Blades in two different algebras.
+
+    This is now just a thin wrapper around :func:`LinearMatrix.from_mapping`,
+    which is more powerful.
 
     Examples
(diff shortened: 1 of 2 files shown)


arsenovic:
so, pedagogically i dont think it adds anything to STA. however, i think its good because it teachs the transformation tools  at the same time, and these are more generally useful

arsenovic:
i like this change.

PR #307, opened 2020-04-09
Add a brief tutorial on using `clifford.transformations`, at Alex's suggestion

eric-wieser:


eric-wieser:
https://clifford--307.org.readthedocs.build/en/307/tutorials/linear-transformations.html

eric-wieser:
Merging as is, but please feel free to make suggestions for extra things of value that could be in the tutorial.

PR #308, opened 2020-04-13
Pin numba below 0.50.0, which we know is a breaking change

eric-wieser:


eric-wieser:
322 warnings against numba 0.49.0rc2 when I run locally with `py -3.8 -m pytest -m "not veryslow"`, but we still work.

PR #312, opened 2020-05-16
Adding Double Conformal Geometric Algebra stub and tests

hugohadfield:


clifford/test/test_dcga.py
```diff
@@ -0,0 +1,243 @@
...
+
+class GeometricPrimitiveTests(unittest.TestCase):
+    def test_reciprocality(self):
+        """
+        Ensure that the cyclide ops and the reciprocal frame are
+        actually reciprocal...
+        """
+        for key, cyc_op in cyclide_ops.items():
+            for key2, cyc_op_recip in cyclide_ops_reciprocal.items():
+                assert cyc_op|cyc_op_recip == (1 + 0*e1)*(key == key2)
```
eric-wieser, line 63:
```suggestion
                assert cyc_op|cyc_op_recip == layout.scalar*(key == key2)
```

clifford/dcga.py
```diff
@@ -0,0 +1,123 @@
+"""
+The Cl(8,2) Double Conformal Geometric Algebra
```
eric-wieser, line 2: Should this file be `dg3c`, since it's the 3d one?

clifford/dcga.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+# The two pseudo-scalars, infinities and origins
+IC1 = e12345
+IC2 = e678910
+einf1 = e4 + e5
+eo1 = 0.5*(e5 - e4)
+einf2 = e9 + e10
+eo2 = 0.5*(e10 - e9)
+
```
eric-wieser, line 30: Would it make sense to call the base blades `ea1, ea2, ...` and `eb1, eb2, ...`?

clifford/test/test_dcga.py
```diff
@@ -0,0 +1,243 @@
...
+        Lcga2 = IC2 * (up_cga2(pnt_vec) ^ up_cga2(pnt_vec + direction_vec) ^ einf2)
+        Ldcga = Lcga1 ^ Lcga2
+
+        # Make a dcga translation rotor in direction of the line
+        Tc1 = 1 - (direction_vec[0] * e1 + direction_vec[1] * e2 + direction_vec[2] * e3) * einf1
+        Tc2 = 1 - (direction_vec[0] * e6 + direction_vec[1] * e7 + direction_vec[2] * e8) * einf2
+        Tdcga = (Tc1*Tc2).normal()
+
+        # Assert the rotor is normalised
+        assert Tdcga*~Tdcga == 1.0 + 0*eo
```
eric-wieser, line 117:
```suggestion
        assert Tdcga*~Tdcga == layout.scalar
```

clifford/dcga.py
hugohadfield: Maybe, I have stuck with this way because it is similar to the way it is written in the paper:
https://link.springer.com/content/pdf/10.1007/s00006-017-0784-0.pdf
We could deviate from this for readability though

eric-wieser: Matching the paper seems like a good enough argument to me.

clifford/dcga.py
eric-wieser: Or `g3dc`

hugohadfield: I don't know what the best thing to do here is.. mostly because I don't really like g3c as a name but nonetheless we already have it. I think maybe dg3c is the way to go

docs/predefined-algebras.rst
```diff
@@ -27,9 +27,9 @@ The easiest way to get started with ``clifford`` is to use one of several predef
 * .. module:: clifford.dpga
 
   ``dpga``: Double PGA also referred to as the Mother Algebra, ``Cl(4, 4)``.
-* .. module:: clifford.dcga
+* .. module:: clifford.dg3c
 
-  ``dcga``: Double Conformal Geometric Algebra (DCGA), ``Cl(8, 2)``.
+  ``dg3c``: Double Conformal Geometric Algebra, ``Cl(8, 2)``.
```
eric-wieser, line 32:
```suggestion
  ``dg3c``: Double Conformal space for G3, ``Cl(8, 2)``.
```
would match the way the other modules express this. Any wording including "3" is fine by me.

This is perhaps an argument for `g3dc` instead of `dg3c`. It also has the nice property that `import clifford.g<tab>` will reveal it.

hugohadfield: The problem here is that multiple papers refer to it as Double Conformal Geometric Algebra or even just by the acronym DCGA

hugohadfield: Also I don't think Double Conformal space for G3 directly captures the concept that this really is two CGAs glued together

hugohadfield: How about?
Double Conformal Geometric Algebra, effectively two g3c algebras glued together

eric-wieser: Last one sounds good enough, thanks.

clifford/dg3c.py
```diff
@@ -0,0 +1,123 @@
...
+eo = eo1^eo2
+einf = einf1^einf2
+
+
+def up_cga1(pnt_vector):
+    """
+    Take a vector and embed it as a point in the first
+    copy of cga
+    """
+    euc_point = pnt_vector[0]*e1 + pnt_vector[1]*e2 + pnt_vector[2]*e3
```
eric-wieser, line 41:
```suggestion
    x, y, z = pnt_vector
    euc_point = x*e1 + y*e2 + z*e3
```

PR #314, opened 2020-05-20
Avoid crashes building llvmlite in CI

eric-wieser:
The later versions of llvmlite do not have a Python 3.5 wheel on PyPI, so travis attempts to build them from source.
This fails unless a suitable local version of llvm is available.

It's easier just to tell `pip` to prefer pre-compiled dependencies.

Caused by https://github.com/numba/llvmlite/issues/471


eric-wieser:
Python 3.5 CI is running again, so we're back to the regular CI whack-a-mole.

Issue #315, opened 2020-05-20
dg3c and dpga tests add almost half an hour to CI


eric-wieser:
From the no-jit run:
```
========================== slowest 25 test durations ===========================
395.32s call     clifford/test/test_dg3c.py::BasicTests::test_up_down
305.63s call     clifford/test/test_dpga.py::TestBasicDPGA::test_line
138.70s call     clifford/test/test_dpga.py::TestBasicDPGA::test_rotate
107.38s call     clifford/test/test_dg3c.py::BasicTests::test_up_down_cga1
100.45s call     clifford/test/test_dg3c.py::GeometricPrimitiveTests::test_rotation
80.05s call     clifford/test/test_dpga.py::TestBasicDPGA::test_translate
52.18s call     clifford/test/test_dg3c.py::GeometricPrimitiveTests::test_bivector_orthogonality
44.94s call     clifford/test/test_dpga.py::TestBasicDPGA::test_up_down
37.76s call     clifford/test/test_dg3c.py::GeometricPrimitiveTests::test_translation
32.41s call     clifford/test/test_dg3c.py::GeometricPrimitiveTests::test_reciprocality
25.90s call     clifford/test/test_dg3c.py::BasicTests::test_metric
19.54s call     clifford/test/test_dg3c.py::GeometricPrimitiveTests::test_line
10.66s call     clifford/test/test_tools.py::G3ToolsTests::test_quaternion_conversions
8.79s call     clifford/test/test_tools.py::G3ToolsTests::test_generate_rotation_rotor_and_angle
8.48s call     clifford/test/test_clifford.py::TestBasicAlgebra::test_right_multiplication_matrix[conformal Cl(3)]
8.21s call     clifford/test/test_clifford.py::TestBasicAlgebra::test_left_multiplication_matrix[conformal Cl(3)]
6.08s call     clifford/test/test_io.py::TestHDF5BasicIO::test_write_and_read
6.00s call     clifford/test/test_io.py::TestJSONBasicIO::test_write_and_read
5.97s call     clifford/test/test_io.py::TestHDF5BasicIO::test_write_and_read_array
4.78s call     clifford/test/test_aliases.py::test_import[gac]
4.73s call     clifford/test/test_g3c_tools.py::TestVisualisation::test_draw_objects
2.81s call     clifford/test/test_dpga.py::TestBasicDPGA::test_bivector_identities
[...]

PR #316, opened 2020-05-20
Reduce the no-jit run time of dpga and dg3c tests

hugohadfield:
Reduced the loop count for all the dpga and dg3c tests


hugohadfield:
See https://github.com/pygae/clifford/issues/315

PR #319, opened 2020-05-28
Delay imports for the large algebra tests

eric-wieser:
This prevents the algebras being constructed when pytest starts up.


eric-wieser:
Coverage saves the day, accidentally disabled all the DPGA tests.

PR #322, opened 2020-05-29
Revert "Disable epub and PDF builds - no one needs them, and they make the build 3x slower"

eric-wieser:
This reverts commit fae278101436b1a614e8756e2a0555c8dcbe991e.

ReadTheDocs is now configured to use a single sphinx build to generate all three outputs.
As a result, the notebooks should now be run just once.


eric-wieser:
See this can't be tested in a PR, let's merge it and see what happens

PR #326, opened 2020-06-05
Experimental AST rewriter and JIT decorator

hugohadfield:


clifford/jit_func.py
```diff
@@ -0,0 +1,64 @@
...
+        # Set things up into locals and globals so that they JIT ok...
+        locals_dict = {'as_ga': self.layout.as_ga_value_vector_func,
+                       'ga_add': self.layout.overload_add_func,
+                       'ga_sub': self.layout.overload_sub_func,
+                       'ga_mul': self.layout.overload_mul_func,
+                       'ga_xor': self.layout.overload_xor_func,
+                       'ga_or': self.layout.overload_or_func}
+        globs = globals()
+        for k, v in locals_dict.items():
+            globs[k] = v
```
eric-wieser, line 51: This updates the global globals, which is unlikely what you want to do.

This might work:

```suggestion
        globs = collections.ChainMap(locals_dict, globals())
```

clifford/_ast_transformer.py
```diff
@@ -0,0 +1,54 @@
...
+            new_node = GATransformer().visit(new_node)
+            return new_node
+        elif isinstance(node.op, ast.Sub):
+            new_node = ast.Call(
+                func=ast.Name(id='ga_sub', ctx=ast.Load()),
+                args=[node.left, node.right],
+                keywords=[]
+            )
+            new_node = GATransformer().visit(new_node)
+            return new_node
```
eric-wieser, line 52: Easy deduplication
```suggestion
        ops = {
            ast.Mult: 'ga_mul',
            ast.BitXor: 'ga_xor',
            ast.BitOr: 'ga_or',
            ast.Add: 'ga_add',
            ast.Sub: 'ga_sub',
        }
        try:
            func_name = ops[type(node.op)]
        except KeyError:
            return node
        else:
            return ast.Call(
                func=ast.Name(id=func_name, ctx=ast.Load()),
                args=[self.visit(node.left), self.visit(node.right)],
                keywords=[]
            )
```

clifford/_layout.py
```diff
@@ -175,6 +179,177 @@ def construct_graded_mt(
...
+    """
+    scalar_index = layout._basis_blade_order.bitmap_to_index[0]
+
+    def ga_sub(x):
+        # dummy function to overload
+        pass
+
+    @overload(ga_sub, inline='always')
+    def ol_ga_sub(a, b):
+        if isinstance(a, (types.Integer, types.Float)) and isinstance(b, types.Array):
```
eric-wieser, line 243:
```suggestion
        if isinstance(a, types.abstract.Number) and isinstance(b, types.Array):
```
throughout, probably.

clifford/test/test_jit_func.py
```diff
@@ -0,0 +1,75 @@
+import unittest
+import numpy as np
+import time
+from ..jit_func import jit_func
+
+
+class TestJITFunc(unittest.TestCase):
```
eric-wieser, line 7:
```suggestion
class TestJITFunc:
```
Gets in the way of pytest

clifford/test/test_jit_func.py
```diff
@@ -0,0 +1,75 @@
...
+        nrepeats = 10000
+        start_time = time.time()
+        for i in range(nrepeats):
+            slow_test_func(e1, e2, einf)
+        end_time = time.time()
+        print('Without jit_func (us): ', 1E6 * (end_time - start_time) / nrepeats)
+
+
+if __name__ == '__main__':
+    unittest.main()
```
eric-wieser, line 75:
```suggestion
```
You can run this file with `pytest clifford/test/test_jit_func.py`.

clifford/test/test_jit_func.py
```diff
@@ -0,0 +1,75 @@
...
+        einf = layout.einf
+
+        @jit_func(layout=layout)
+        def test_func(A, B, C):
+            op = ~(((A * B) * C) | (B ^ A)) - 3.1 - A - 7 * B + 5 + C + 2.5 + (2 ^ (A * B * C) ^ 3) + (A | 5)
+            return op
+
+        def slow_test_func(A, B, C):
+            op = ~(((A * B) * C) | (B ^ A)) - 3.1 - A - 7 * B + 5 + C + 2.5 + (2 ^ (A * B * C) ^ 3) + (A | 5)
+            return op
```
eric-wieser, line 54: To avoid repeating the function:
```suggestion
        def slow_test_func(A, B, C):
            op = ~(((A * B) * C) | (B ^ A)) - 3.1 - A - 7 * B + 5 + C + 2.5 + (2 ^ (A * B * C) ^ 3) + (A | 5)
            return op
            
        test_func = jit_func(layout)(slow_test_func)
```

clifford/test/test_jit_func.py
```diff
@@ -0,0 +1,75 @@
...
+
+        np.testing.assert_allclose(test_func(e1, e2, einf).value, slow_test_func(e1, e2, einf).value)
+
+    def test_benchmark(self):
```
eric-wieser, line 41:
```suggestion
    @pytest.mark.parametrize('use_jit', [False, True]
    def test_benchmark(self, use_jit, benchmark):
```

clifford/test/test_jit_func.py
```diff
@@ -0,0 +1,75 @@
...
+            test_func(e1, e2, einf)
+        end_time = time.time()
+        print('With jit_func (us): ', 1E6 * (end_time - start_time) / nrepeats)
+
+        nrepeats = 10000
+        start_time = time.time()
+        for i in range(nrepeats):
+            slow_test_func(e1, e2, einf)
+        end_time = time.time()
+        print('Without jit_func (us): ', 1E6 * (end_time - start_time) / nrepeats)
```
eric-wieser, line 71:
```suggestion
        benchmark(test_func, e1, e2, ninf)
```

clifford/jit_func.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(self, layout, ast_debug=False):
+        self.layout = layout
+        if AST_PRETTY_AVAILABLE:
+            self.ast_debug = ast_debug
+        else:
+            if ast_debug:
+                warnings.warn('''
+The ast_debug flag is set to True, but the astpretty module is not importable.
+To see ast_debug output please pip install astpretty
+''')
```
eric-wieser, line 28:
```suggestion
                warnings.warn(
                    'The ast_debug flag is set to True, but the astpretty module is not importable.\n'
                    'To see ast_debug output please pip install astpretty')
```
or `textwrap.dedent`, but that seems overkill for two lines of text.

clifford/_layout.py
```diff
@@ -175,6 +179,201 @@ def construct_graded_mt(
...
+                return omt_func(a, as_ga(b))
+            return impl
+        elif isinstance(a, types.abstract.Number) and isinstance(b, types.Array):
+            def impl(a, b):
+                return omt_func(as_ga(a), b)
+            return impl
+        else:
+            def impl(a, b):
+                return a^b
+            return impl
```
eric-wieser, line 315: Tempting to make these shorter:
```suggestion
            return lambda a, b: omt_func(a, b)
        elif isinstance(a, types.Array) and isinstance(b, types.abstract.Number):
            return lambda a, b: omt_func(a, as_ga(b))
        elif isinstance(a, types.abstract.Number) and isinstance(b, types.Array):
            return lambda a, b: omt_func(as_ga(a), b)
        else:
            return lambda a, b: return a^b
```

clifford/jit_func.py
```diff
@@ -0,0 +1,80 @@
+
+import ast
+try:
+    import astpretty
+    AST_PRETTY_AVAILABLE = True
+except:
```
eric-wieser, line 6:
```suggestion
except ImportError:
```

clifford/_ast_transformer.py
```diff
@@ -0,0 +1,59 @@
...
+    def visit_Call(self, node):
+        try:
+            nfuncid = node.func.id
+            return node
+        except:
+            # Only allow a single grade to be selected for now
+            if len(node.args) == 1:
+                return ast.Call(
+                    func=ast.Name(id='ga_call', ctx=ast.Load()),
+                    args=[node.func, node.args[0]],
```
eric-wieser, line 55:
```suggestion
                    args=[self.visit(node.func), node.args[0]],
```

eric-wieser: Otherwise `(a *b)(0)` will not work

clifford/test/test_jit_func.py
hugohadfield: I had a go at using this syntax and it really wasn't happy, I think calling the decorator in this way messes up the link to the source in some way, didn't dig into it too deeply

clifford/jit_func.py
```diff
@@ -0,0 +1,80 @@
...
+                    'The ast_debug flag is set to True, but the astpretty module is not importable.\n'
+                    'To see ast_debug output please pip install astpretty')
+            self.ast_debug = False
+
+    def __call__(self, func):
+        # Get the function source
+        fname = func.__name__
+        source = inspect.getsource(func)
+        # Remove the decorator first line.
+        source = '\n'.join(source.splitlines()[1:])
```
eric-wieser, line 35: It would be better to remove this at the ast level

clifford/test/test_jit_func.py
eric-wieser: I suspect it's just you trimming a decorator which is no longer there

hugohadfield: that's a fair point actually, I guess changing the trim to be in the ast would probably remove that problem, I'll have a look at it

clifford/test/test_jit_func.py
```diff
@@ -23,6 +23,31 @@ def slow_test_func(A, B):
...
+        def compound_func(A, B):
+            return test_func_2(test_func_1(A, B))
+
+        test_func_1_jit = jit_func(self.layout)(test_func_1)
+        test_func_2_jit = jit_func(self.layout,
+                                   mv_constants={'e12': e12})(test_func_2)
+
+        test_compound_func = jit_func(self.layout,
+                                      mv_constants={'test_func_1': test_func_1_jit,
+                                                    'test_func_2': test_func_2_jit})(compound_func)
```
hugohadfield, line 47: This is a way to nest functions by using(abusing?) the mv_constants argument, it is clearly not a great solution but I'm not sure how to pull it off otherwise. Perhaps the decorator has knowledge of what functions are available?

clifford/_ast_transformer.py
```diff
@@ -0,0 +1,63 @@
+
+import ast
+
+
+class DecoratorRemover(ast.NodeTransformer):
+    """ Strip decorators from FunctionDefs"""
```
eric-wieser, line 6:
```suggestion
    """ Strip decorators from top-level FunctionDefs"""
```

This doesn't recurse, but that's a good thing

clifford/_ast_transformer.py
```diff
@@ -0,0 +1,63 @@
...
+        if len(node.args) == 1:
+            return ast.Call(
+                func=ast.Name(id='ga_call', ctx=ast.Load()),
+                args=[self.visit(node.func), self.visit(node.args[0])],
+                keywords=[]
+            )
+        else:
+            node.func = self.visit(node.func)
+            node.args = [self.visit(a) for a in node.args]
+            return node
```
eric-wieser, line 63: This misses kwargs:
```suggestion
            return self.generic_visit(node)
```

clifford/_ast_transformer.py
```diff
@@ -0,0 +1,63 @@
...
+                keywords=[]
+            )
+
+    def visit_Call(self, node):
+        if len(node.args) == 1:
+            return ast.Call(
+                func=ast.Name(id='ga_call', ctx=ast.Load()),
+                args=[self.visit(node.func), self.visit(node.args[0])],
+                keywords=[]
+            )
```
eric-wieser, line 59: Does the wrong thing for `f(mv, kwarg="lost")`. Better as:
```suggestion
            node = self.generic_visit(node)
            node.args = [node.func] + node.args
            node.func = ast.Name(id='ga_call', ctx=ast.Load())
            return node
```

clifford/test/test_jit_func.py
eric-wieser: The decorator can probably use `inspect.currentframe().f_back.f_locals`

clifford/_layout.py
```diff
@@ -596,6 +844,38 @@ def left_complement_func(self):
...
+    def overload_reverse_func(self):
+        return get_overload_reverse(self)
+
+    @_cached_property
+    def project_to_grade_func(self):
+        return get_project_to_grade_func(self)
+
+    @_cached_property
+    def overload_call_func(self):
+        return get_overload_call(self)
```
eric-wieser, line 877: I don't think these really belong here. You've put them here because it makes them easy to cache, but I think it would be better to cache them manually in a `weakref.WeakKeyDictionary`.

clifford/jit_func.py
```diff
@@ -0,0 +1,92 @@
+
```
eric-wieser, line 1: Perhaps `clifford/experimental/jit_func`

clifford/_layout.py
eric-wieser: So perhaps something like
```
def weak_cache(f):
    _cache = weakref.WeakKeyDictionary()
    @functools.wraps(f)
    def wrapped(*args, **kwargs):
        a, *args = args
        try:
            return _cache[a]
        except KeyError:
            ret =_cache[a] = f(a, *args, **kwargs)
            return ret
    wrapped._cache = _cache
    return wrapped


@weak_cache
def _get_jit_impls(layout):
    return {
        'as_ga': get_as_ga_value_vector_func(layout),
        'ga_add': get_overload_add_func(layout),
        'ga_sub': get_overload_sub_func(layout),
        'ga_mul': get_overload_mul_func(layout),
        'ga_xor': get_overload_xor_func(layout),
        'ga_or': get_overload_or_func(layout),
        'ga_rev': get_overload_reverse_func(layout),
        'ga_call': get_overload_call_func(layout),
    }
```

clifford/jit_func.py
```diff
@@ -0,0 +1,92 @@
...
+        source = textwrap.dedent(source)
+
+        # Parse the source
+        tree = ast.parse(source)
+        if self.ast_debug:
+            print('\n\n\n\n TRANSFORMING FROM \n\n\n\n')
+            astpretty.pprint(tree)
+
+        # Remove the decorators from the function
+        tree = DecoratorRemover().visit(tree)
```
eric-wieser, line 47: Strictly you want to leave behind all the non-`jitfunc` decorators, but that's fine as a todo.

clifford/jit_func.py
```diff
@@ -0,0 +1,92 @@
...
+                       'ga_xor': self.layout.overload_xor_func,
+                       'ga_or': self.layout.overload_or_func,
+                       'ga_rev': self.layout.overload_reverse_func,
+                       'ga_call': self.layout.overload_call_func}
+
+        # Add the passed multivector and scalar constants
+        for k, v in self.mv_constants.items():
+            locals_dict[k] = v.value
+        for k, v in self.scalar_constants.items():
+            locals_dict[k] = v
```
eric-wieser, line 71:
```suggestion
        # Add the passed multivector and scalar constants
        locals_dict.update(self.mv_constants)
        locals_dict.update(self.scalar_constants)
```

clifford/jit_func.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # TODO: Work out a better way to deal with changes to globals
+        globs = {}
+        for k, v in globals().items():
+            globs[k] = v
+        for k, v in locals_dict.items():
+            globs[k] = v
```
eric-wieser, line 78:
```suggestion
        globs.update(globals)
        globs.update(locals_dict)
```

clifford/test/test_jit_func.py
eric-wieser: I suspect you should be using `f_globals` instead of `globals()` too.

clifford/__init__.py
```diff
@@ -117,7 +117,8 @@ def linear_operator_as_matrix(func, input_blades, output_blades):
     ndimout = len(output_blades)
     mat = np.zeros((ndimout, ndimin))
     for i, b in enumerate(input_blades):
-        mat[:, i] = np.array([func(b)[j] for j in output_blades])
```
eric-wieser, line 120: Nice find

eric-wieser: Probably makes sense to spin out a quick PR with just this fix against master.

hugohadfield: yeah I was thinking that

eric-wieser:
What do we want to do with this now that we have proper numba support?

hugohadfield:
Shall we just close it? I don't think it adds anything on top of what we have, although it was an interesting and fun idea

PR #329, opened 2020-06-10
tools.g3c: Invert the relationship between `func` and `val_func` functions

eric-wieser:
This changes all the algorithms to be implemented using the numba overloading syntax.
This isn't exhaustive, but hits most of the cases.

We can't remove the `val_func` functions until we remove all the callers.

This will probably need quite careful review, as I don't think all these functions are tested. I'll make sure not to do any rebases so that it's easy to keep track of where you get to.

### DIFF
--- a/clifford/tools/g3c/__init__.py
+++ b/clifford/tools/g3c/__init__.py
@@ -214,6 +214,11 @@ mask5 = layout.get_grade_projection_matrix(5)
 mask_2minus4 = mask2 - mask4
 
 
+def _defunct_wrapper(f):
+    """ marker for val_ functions which are no longer faster """
+    return f
+
+
 def interpret_multivector_as_object(mv):
     """
     Takes an input multivector and returns what kind of object it is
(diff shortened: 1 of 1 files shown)


clifford/tools/g3c/__init__.py
```diff
@@ -1808,43 +1782,53 @@ def convert_2D_polar_line_to_conformal_line(rho, theta):
...
-    return mv_val - no.value + omt_func(temp, gmt_func(gmt_func(mv_val, mv_val), ninf.value))
+    return fast_up(layout.MultiVector(mv_val)).value
 
 
 @numba.njit
 def fast_up(mv):
-    """ Fast up mapping """
-    return layout.MultiVector(val_up(mv.value))
+    """ Fast jitted up mapping """
+    return mv - no + (0.5 * ((mv * mv) * ninf))
```
hugohadfield, line 1793: big wins for readability!

PR #336, opened 2020-06-19
numba: Add support for the projection operator

eric-wieser:
Numba does not make it easy to overload `__call__` at the moment.
We should follow up with numba to see if they can make it easier in future.

xref https://github.com/numba/numba/issues/5885.

### DIFF
--- a/clifford/tools/g3c/__init__.py
+++ b/clifford/tools/g3c/__init__.py
@@ -674,7 +674,7 @@ def get_line_intersection(L3, Ldd):
     P = -(Pd * ninf * Pd)
     imt = Pd | ninf
     P_denominator = 2*(imt * imt).value[0]
-    return _project(P/P_denominator, 1)
+    return (P/P_denominator)(1)
 
 
 @numba.njit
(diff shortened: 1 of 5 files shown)


clifford/tools/g3c/__init__.py
```diff
@@ -854,26 +854,7 @@ def random_rotation_translation_rotor(maximum_translation=10.0, maximum_angle=np
...
-        output[1:6] = mv.value[1:6]
-    elif grade == 2:
-        output[6:16] = mv.value[6:16]
-    elif grade == 3:
-        output[16:26] = mv.value[16:26]
-    elif grade == 4:
-        output[26:31] = mv.value[26:31]
-    elif grade == 5:
-        output[31] = mv.value[31]
-    return mv.layout.MultiVector(output)
```
eric-wieser, line 876: This doesn't actually seem to be any faster than my approach.

eric-wieser:
3.8 CI is the usual, LGTM complaints are https://github.com/github/codeql/issues/3741 (invalid) and don't-care.

PR #337, opened 2020-06-19
numba: Add support for abs(mv) and mv.norm

eric-wieser:
This makes the `norm` and `normalised` functions from tools.g3c unnecessary.
Conveniently, we can mostly reuse the pure-python implementations here, as these functions are trivial.

### DIFF
--- a/clifford/numba/_multivector.py
+++ b/clifford/numba/_multivector.py
@@ -306,3 +306,21 @@ def ga_call(self, arg):
             mv.value[inds] = self.value[inds]
             return mv
         return impl
+
+
+@numba.extending.overload_method(MultiVectorType, 'mag2')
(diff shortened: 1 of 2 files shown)


clifford/numba/_multivector.py
```diff
@@ -306,3 +306,21 @@ def impl(self, arg):
             mv.value[inds] = self.value[inds]
             return mv
         return impl
+
+
+@numba.extending.overload_method(MultiVectorType, 'mag2')
+def MultiVector_mag2(self):
+    def impl(self):
+        return (~self * self).value[0]
```
eric-wieser, line 314: This one we can't reuse because of the optimizations made in the pure-python implementation...

PR #342, opened 2020-06-25
docfix: g3 turorial missing pi,e import

arsenovic:
minor fix for rtd

### DIFF
--- a/docs/tutorials/g3-algebra-of-space.ipynb
+++ b/docs/tutorials/g3-algebra-of-space.ipynb
@@ -39,11 +39,12 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
     "import clifford as cf\n",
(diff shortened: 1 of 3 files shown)


docs/tutorials/g3-algebra-of-space.ipynb
```diff
@@ -39,11 +39,12 @@
   },
```
eric-wieser, line 39: <p>I think it would be better to just change this box to:</p><pre class="ql-syntax" spellcheck="false">from math import cos, sin, pi

theta = pi/4
R = cos(theta) - sin(theta)*e23
R
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pygae/clifford/pull/342/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

eric-wieser: (this comment only makes sense on ReviewNB it seems...)

PR #343, opened 2020-06-26
Added the start of ndarray of multivectors

hugohadfield:
Very basic Nd array support

### DIFF
--- a/clifford/_mvarray.py
+++ b/clifford/_mvarray.py
@@ -45,32 +45,31 @@ class MVArray(np.ndarray):
         for index in np.ndindex(input_shape):
             obj[index] = _index_nested_iterable(input_array, index)
 
-        # Record the layout and dtype
         self = obj.view(cls)
-        self.layout = layout
-        self.value_dtype = dtype
-
-        # Set up the vectorised function
-        self.v_value_get = np.vectorize(lambda x: x.value, otypes=[self.value_dtype], signature='()->(n)')
         return self
 
     def __array_finalize__(self, obj):
         if obj is None:
             return
 
+    def _get_first_element(self):
+        return _index_nested_iterable(self, [0] * len(self.shape))
+
     @property
     def value(self):
         """
         Return an np array of the values of multivectors
         """
-        return self.v_value_get(self)
+        value_dtype = self._get_first_element().value.dtype
+        v_value_get = np.vectorize(lambda x: x.value, otypes=[value_dtype], signature='()->(n)')
+        return v_value_get(self)
 
     @staticmethod
     def from_value_array(layout, value_array):
         """
         Constructs an array of mvs from a value array
         """
-        v_new_mv = np.vectorize(lambda v: MultiVector(layout, v), otypes=[MVArray], signature='(n)->()')
+        v_new_mv = np.vectorize(lambda v: MultiVector(layout, v), otypes=[MultiVector], signature='(n)->()')
         return MVArray(v_new_mv(value_array))
 
     def save(self, filename, compression=True, transpose=False,
@@ -78,7 +77,9 @@ class MVArray(np.ndarray):
         """
         Saves the array to a ga file
         """
-        write_ga_file(filename, self.value, self[0].layout.metric, self[0].layout.basis_names,
+        first_element = self._get_first_element()
+        write_ga_file(filename, self.value, first_element.layout.metric,
+                      first_element.layout.basis_names,
                       compression=compression, transpose=transpose,
                       sparse=sparse, support=support, compression_opts=compression_opts)
 
--- a/clifford/_mvarray.py
+++ b/clifford/_mvarray.py
@@ -38,12 +38,21 @@ class MVArray(np.ndarray):
     MultiVector Array
     """
     def __new__(cls, input_array):
+
         input_shape, layout, dtype = _interrogate_nested_mvs(input_array)
+        # copying this across elementwise is necessary to prevent numpy recursing into the multivector coefficients
         obj = np.empty(input_shape, dtype=object)
         for index in np.ndindex(input_shape):
             obj[index] = _index_nested_iterable(input_array, index)
-        obj = obj.view(cls)
-        return obj
+
+        # Record the layout and dtype
+        self = obj.view(cls)
+        self.layout = layout
+        self.value_dtype = dtype
+
+        # Set up the vectorised function
+        self.v_value_get = np.vectorize(lambda x: x.value, otypes=[self.value_dtype], signature='()->(n)')
+        return self
 
     def __array_finalize__(self, obj):
         if obj is None:
@@ -54,8 +63,7 @@ class MVArray(np.ndarray):
         """
         Return an np array of the values of multivectors
         """
-        v_value_get = np.vectorize(lambda x: x.value, otypes=[np.float], signature='()->(n)')
-        return v_value_get(self)
+        return self.v_value_get(self)
 
     @staticmethod
     def from_value_array(layout, value_array):
--- a/clifford/_mvarray.py
+++ b/clifford/_mvarray.py
@@ -1,11 +1,11 @@
+import functools
+import operator
+from typing import Tuple
+
 import numpy as np
 
-import operator
-import functools
-
-from typing import Tuple
-from clifford import Layout
 from clifford.io import write_ga_file, read_ga_file  # noqa: F401
+from ._layout import Layout
 from ._multivector import MultiVector
 
 dual_array = np.vectorize(MultiVector.dual)
--- a/clifford/_mvarray.py
+++ b/clifford/_mvarray.py
@@ -1,6 +1,10 @@
 import numpy as np
 
+import operator
+import functools
 
+from typing import Tuple
+from clifford import Layout
 from clifford.io import write_ga_file, read_ga_file  # noqa: F401
 from ._multivector import MultiVector
 
@@ -9,16 +13,16 @@ normal_array = np.vectorize(MultiVector.normal)
 call_array = np.vectorize(MultiVector.__call__)
 
 
-def _interrogate_nested_mvs(input_array):
+def _interrogate_nested_mvs(input_array) -> Tuple[Tuple[int], Layout, np.dtype]:
     """
     Calculates the shape of the nested input_array, and gets the associated layout.
     Stops descending when it encounters a MultiVector.
     """
-    if not isinstance(input_array[0], MultiVector):
+    if not isinstance(input_array, MultiVector):
         nested_shape, layout, dtype = _interrogate_nested_mvs(input_array[0])
         return (len(input_array), *nested_shape), layout, dtype
     else:
-        return tuple([len(input_array)]), input_array[0].layout, type(input_array[0].value[0])
+        return (), input_array.layout, input_array.value.dtype
 
 
 def _index_nested_iterable(input_iterable, index):
@@ -26,10 +30,7 @@ def _index_nested_iterable(input_iterable, index):
     Given a nested iterable, input_iterable, return the element given by the
     1d index iterable
     """
-    res = input_iterable[index[0]]
-    for ind in index[1:]:
-        res = res[ind]
-    return res
+    return functools.reduce(operator.getitem, index, input_iterable)
 
 
 class MVArray(np.ndarray):
--- a/clifford/_mvarray.py
+++ b/clifford/_mvarray.py
@@ -33,7 +33,7 @@ def _index_nested_iterable(input_iterable, index):
     return functools.reduce(operator.getitem, index, input_iterable)
 
 
-@functools.lru_cache
+@functools.lru_cache(None)
 def _get_vectorized_value_func(dtype):
     return np.vectorize(lambda x: x.value, otypes=[dtype], signature='()->(n)')
 
--- a/clifford/_mvarray.py
+++ b/clifford/_mvarray.py
@@ -58,7 +58,7 @@ class MVArray(np.ndarray):
             return
 
     def _get_first_element(self):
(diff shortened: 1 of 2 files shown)


clifford/_mvarray.py
```diff
@@ -9,14 +9,38 @@
...
-
+    """
     def __new__(cls, input_array):
-        obj = np.empty(len(input_array), dtype=object)
-        obj[:] = input_array
+        input_shape, layout, dtype = _interrogate_nested_mvs(input_array)
+        obj = np.empty(input_shape, dtype=object)
+        for index in np.ndindex(input_shape):
+            obj[index] = _index_nested_iterable(input_array, index)
         obj = obj.view(cls)
```
eric-wieser, line 44: In my opinion `dtype=object` is a dead end, and there is little point pursuing it.

eric-wieser: Although I suppose this patch is still reasonable

hugohadfield: Agreed on this in the long term, would be much better to have something that dealt with the mv arrays properly. As an easy stopgap though the dtype=object is fairly convenient

clifford/_mvarray.py
```diff
@@ -9,14 +9,38 @@
...
+def _interrogate_nested_mvs(input_array):
+    """
+    Calculates the shape of the nested input_array, and gets the associated layout.
+    Stops descending when it encounters a MultiVector.
+    """
+    if not isinstance(input_array[0], MultiVector):
+        nested_shape, layout, dtype = _interrogate_nested_mvs(input_array[0])
+        return (len(input_array), *nested_shape), layout, dtype
+    else:
+        return tuple([len(input_array)]), input_array[0].layout, type(input_array[0].value[0])
```
eric-wieser, line 21:
```suggestion
        first_mv = input_array[0]
        return (len(input_array),), first_mv.layout, first_mv.value.dtype
```

clifford/_mvarray.py
```diff
@@ -9,14 +9,38 @@
 call_array = np.vectorize(MultiVector.__call__)
 
 
+def _interrogate_nested_mvs(input_array):
```
eric-wieser, line 12:
```suggestion
def _interrogate_nested_mvs(input_array) -> Tuple[Tuple[int], Layout, np.dtype]:
```
You'll need to import `Layout` too.

clifford/_mvarray.py
eric-wieser, line 21: Your recursion can have an even simpler base case:
```suggestion
    if not isinstance(input_array, MultiVector):
        nested_shape, layout, dtype = _interrogate_nested_mvs(input_array)
        return (len(input_array), *nested_shape), layout, dtype
    else:
        return (), input_array.layout, input_array.value.dtype
```

clifford/_mvarray.py
```diff
@@ -9,14 +9,38 @@
...
+
+def _index_nested_iterable(input_iterable, index):
+    """
+    Given a nested iterable, input_iterable, return the element given by the
+    1d index iterable
+    """
+    res = input_iterable[index[0]]
+    for ind in index[1:]:
+        res = res[ind]
+    return res
```
eric-wieser, line 32: Again, can move the base-case back:
```suggestion
    res = input_iterable
    for ind in index:
        res = res[ind]
    return res
```

clifford/_mvarray.py
```diff
@@ -9,14 +9,38 @@
...
+    """
     MultiVector Array
-    '''
-
+    """
     def __new__(cls, input_array):
-        obj = np.empty(len(input_array), dtype=object)
-        obj[:] = input_array
+        input_shape, layout, dtype = _interrogate_nested_mvs(input_array)
+        obj = np.empty(input_shape, dtype=object)
```
eric-wieser, line 41:
```suggestion
        # copying this across elementwise is necessary to prevent numpy recursing into the multivector coefficients
        obj = np.empty(input_shape, dtype=object)
```

clifford/_mvarray.py
```diff
@@ -29,7 +53,8 @@ def value(self):
         """
         Return an np array of the values of multivectors
         """
-        return np.array([mv.value for mv in self])
+        v_value_get = np.vectorize(lambda x: x.value, otypes=[np.ndarray], signature='()->(n)')
```
eric-wieser, line 56: You should vectorize this in the constructor.

Also, this `otype` is wrong - can you use `float`? to avoid the cast below?

eric-wieser: Better than `float` would be to use the `dtype` returned by `_interrogate_nested_mvs`

clifford/_mvarray.py
```diff
@@ -1,6 +1,10 @@
 import numpy as np
 
+import operator
+import functools
 
+from typing import Tuple
+from clifford import Layout
 from clifford.io import write_ga_file, read_ga_file  # noqa: F401
 from ._multivector import MultiVector
```
eric-wieser, line 9:
```suggestion
from clifford.io import write_ga_file, read_ga_file  # noqa: F401
from ._layout import Layout
from ._multivector import MultiVector
```
makes it easier to reason about circular imports

clifford/_mvarray.py
```diff
@@ -1,6 +1,10 @@
 import numpy as np
 
+import operator
+import functools
 
+from typing import Tuple
```
eric-wieser, line 6: Style is usually to put stdlib imports first:
```suggestion
import functools
import operator
from typing import Tuple

import numpy as np
```

clifford/test/test_clifford.py
```diff
@@ -196,6 +196,55 @@ def test_add_float64(self, g3):
...
+
+        value_array_a = np.zeros((Nrows, Ncolumns, layout.gaDims))
+        for i in range(Nrows):
+            for j in range(Ncolumns):
+                value_array_a[i, j, :] = layout.randomMV().value
+
+        value_array_b = np.zeros((Nrows, Ncolumns, layout.gaDims))
+        for i in range(Nrows):
+            for j in range(Ncolumns):
+                value_array_b[i, j, :] = layout.randomMV().value
```
eric-wieser, line 212: Nit: helper function makes this less duplicated:
```suggestion
    def _random_value_array(self, layout, Nrows, Ncolumns):
        value_array = np.zeros((Nrows, Ncolumns, layout.gaDims))
        for i in range(Nrows):
            for j in range(Ncolumns):
                value_array[i, j, :] = layout.randomMV().value
        return value_array

    def test_2d_mv_array(self, g3):
        layout, blades = g3, g3.blades
        Nrows = 2
        Ncolumns = 3

        value_array_a = self._random_value_array(g3, Nrows, Ncolumns)
        value_array_b = self._random_value_array(g3, Nrows, Ncolumns)
```

clifford/_mvarray.py
hugohadfield: agreed on the dtype, but I was struggling to work out how to get hold of it properly in the MVArray instance. __new__ appears to operate on the class `cls`... can I just write `cls.dtype = dtype` or will that modify it for all instances of MVArray?

eric-wieser: `self` is the return value of `.view` - you can attach properties to that object.

Don't use `.dtype` though, as that's managed by numpy!

clifford/_mvarray.py
```diff
@@ -1,22 +1,48 @@
...
     def __new__(cls, input_array):
-        obj = np.empty(len(input_array), dtype=object)
-        obj[:] = input_array
+        input_shape, layout, dtype = _interrogate_nested_mvs(input_array)
+        # copying this across elementwise is necessary to prevent numpy recursing into the multivector coefficients
+        obj = np.empty(input_shape, dtype=object)
+        for index in np.ndindex(input_shape):
+            obj[index] = _index_nested_iterable(input_array, index)
         obj = obj.view(cls)
         return obj
```
eric-wieser, line 47:
```suggestion
        self = obj.view(cls)
        self.layout = layout
        self.value_dtype = dtype
        return obj
```

clifford/_mvarray.py
```diff
@@ -29,7 +55,8 @@ def value(self):
         """
         Return an np array of the values of multivectors
         """
-        return np.array([mv.value for mv in self])
+        v_value_get = np.vectorize(lambda x: x.value, otypes=[np.float], signature='()->(n)')
```
eric-wieser, line 58: `np.float` raises a DeprecationWarning in numpy master. Let's fix this to use `self.layout_dtype`.

hugohadfield:
image

```

self = MVArray([[1.4594 - (1.19656^e1) - (0.72878^e2) - (1.01381^e3) - (0.40842^e12) - (2.07411^e13) + (0.78918^e23) + (0.525...) + (0.15542^e2) + (0.11253^e3) - (1.0553^e12) + (3.57699^e13) - (0.78675^e23) + (1.8036^e123)]],
        dtype=object)

    @property
    def value(self):
        """
        Return an np array of the values of multivectors
        """
>       return self.v_value_get(self)
E       AttributeError: 'MVArray' object has no attribute 'v_value_get'

../_mvarray.py:64: AttributeError
```

clifford/_mvarray.py
```diff
@@ -1,57 +1,96 @@
...
+
+def _index_nested_iterable(input_iterable, index):
+    """
+    Given a nested iterable, input_iterable, return the element given by the
+    1d index iterable
+    """
+    return functools.reduce(operator.getitem, index, input_iterable)
+
+
+@functools.lru_cache
```
eric-wieser, line 36:
```suggestion
@functools.lru_cache(None)
```
Needs an argument for the cache size

clifford/test/test_clifford.py
```diff
@@ -196,6 +196,54 @@ def test_add_float64(self, g3):
...
+        Nrows = 2
+        Ncolumns = 3
+        value_array_a = self._random_value_array(g3, Nrows, Ncolumns)
+        value_array_b = self._random_value_array(g3, Nrows, Ncolumns)
+
+        mv_array_a = MVArray.from_value_array(layout, value_array_a)
+        assert mv_array_a.shape == (Nrows, Ncolumns)
+        mv_array_b = MVArray.from_value_array(layout, value_array_b)
+        assert mv_array_b.shape == (Nrows, Ncolumns)
+
```
eric-wieser, line 217:
```suggestion
# check the array is preserved (no need to check both)
assert_array_equal(mv_array_a.value, value_array_a)
assert mv_array_a.value.dtype == value_array_a
assert type(mv_array_a.value) == type(value_array_a)

```

clifford/_mvarray.py
```diff
@@ -1,57 +1,96 @@
...
+        value_dtype = self._get_first_element().value.dtype
+        return _get_vectorized_value_func(value_dtype)(self)
 
     @staticmethod
     def from_value_array(layout, value_array):
         """
         Constructs an array of mvs from a value array
         """
-        v_new_mv = np.vectorize(lambda v: MultiVector(layout, v), otypes=[MVArray], signature='(n)->()')
+        v_new_mv = np.vectorize(lambda v: layout.MultiVector(v), otypes=[MultiVector], signature='(n)->()')
```
eric-wieser, line 76:
```suggestion
        v_new_mv = np.vectorize(lambda v: layout.MultiVector(v), otypes=[object], signature='(n)->()')
```
`otypes` is expected to be actual dtypes.

clifford/_mvarray.py
```diff
@@ -1,57 +1,96 @@
...
 from clifford.io import write_ga_file, read_ga_file  # noqa: F401
+from ._layout import Layout
 from ._multivector import MultiVector
 
 dual_array = np.vectorize(MultiVector.dual)
 normal_array = np.vectorize(MultiVector.normal)
 call_array = np.vectorize(MultiVector.__call__)
 
 
+def _interrogate_nested_mvs(input_array) -> Tuple[Tuple[int], Layout, np.dtype]:
```
eric-wieser, line 16: Whoops, had this annotation wrong:
```suggestion
def _interrogate_nested_mvs(input_array) -> Tuple[Tuple[int, ...], Layout, np.dtype]:
```

clifford/_mvarray.py
```diff
@@ -1,57 +1,96 @@
...
+        return _index_nested_iterable(self, [0] * len(self.shape))
+
     @property
     def value(self):
         """
         Return an np array of the values of multivectors
         """
-        return np.array([mv.value for mv in self])
+        value_dtype = self._get_first_element().value.dtype
+        return _get_vectorized_value_func(value_dtype)(self)
```
eric-wieser, line 69: If the test I suggest below doesn't pass, this will need to become
```suggestion
        return _get_vectorized_value_func(value_dtype)(self).view(np.ndarray)
```

clifford/test/test_clifford.py
hugohadfield: This produces the error
```
        # check properties of the array are preserved (no need to check both a and b)
        np.testing.assert_array_equal(mv_array_a.value, value_array_a)
>       assert mv_array_a.value.dtype == value_array_a
E       TypeError: data type not understood

test_clifford.py:249: TypeError
```
With or without the change above to add .view(ndarray)

eric-wieser: I missed a `.dtype`, updated.

clifford/_mvarray.py
```diff
@@ -1,57 +1,96 @@
...
+
+        self = obj.view(cls)
+        return self
 
     def __array_finalize__(self, obj):
         if obj is None:
             return
 
+    def _get_first_element(self):
+        return _index_nested_iterable(self, [0] * len(self.shape))
```
eric-wieser, line 61:
```suggestion
        return self[(0,) * self.ndim]
```

Issue #345, opened 2020-06-26
Consider deprecating `MultiVector.__getitem__(int)`


eric-wieser:
Today, `mv[i]` is overloaded to mean `mv.value[i]`. I think we should deprecate this behavior because:

* It is not commonly used. A large fraction of the uses are `mv[0]`, and these uses are actually a bug if used with a `BasisBladeOrder` that does not store the scalar in position 0. These cases should be spelt `mv[()]` to be strictly used
* Where it is used, it is typically an optimization that has intricate knowledge of the storage order of `.value`. In these locations, the burden of using `.value[...]` directly is not high
* It is impossible to generalize to arrays of multivectors, as it becomes indisinguishable from element selection
* It shares responsibility for some of the complaints in gh-344

Issue #344, opened 2020-06-26
Consider deprecating `MultiVector.__len__`


eric-wieser:
Today, `len(mv)` returns `mv.layout.gaDims`. I think we should deprecate this behavior because:

* It's not used frequently enough for the short spelling to be worthwhile
* It doesn't generalize well to treating N-multivectors as a subspace of (N+1)-multivectors
* In conjunction with `__getitem__`, it allows `list(mv)` which isn't mathematically meaningful (it exposes internal information), and is much slower than `mv.value`.
* If the user is using it exactly because they want implementation details, they can just use `len(mv.value)`.
* `len(mv_array)` must have a different and unrelated meaning

hugohadfield:
I support this for sure

arsenovic:
i agree with this.

hugohadfield:
Just found an example of code reliant on len(mv) in the wild, our own pyganja:
https://github.com/pygae/pyganja/blob/19849957dfba5f9f0b558c072fb7dfd783db9f0d/pyganja/GanjaScene.py#L46

hugohadfield:
Hopefully now removed the use of the above via https://github.com/pygae/pyganja/pull/45

PR #346, opened 2020-06-26
Deprecate the sequence-like methods of MultiVector

eric-wieser:
Addresses gh-344, gh-345.

In my opinion it is wrong to say that a multivector _is_ a sequence of coefficients - it happens to _have_ a sequence of coefficients, and if you want to access these you get all the sequence behavior through `.value`.

Let's see how many warnings our tests suite produces.


eric-wieser:
A lot of the warnings stem from the `(a * b)[0]` or `(a | b)[0]` spelling of the scalar product, which we could address in #347.

eric-wieser:
Think I got all but one of the remaining uses of this API - #343 will remove the last one.

PR #353, opened 2020-08-19
Add a mailmap to merge similar email addresses in the git log

eric-wieser:
Not sure this affects github, but it affects the git command line.

Note that this was already done in galgebra.


hugohadfield (approved):
Looks fine to me

Issue #354, opened 2020-09-10
Ensure full support for Clifford algebras over complex numbers


hugohadfield:
Currently the package has accidental support for complex clifford algebras but this has not been an explicit design goal. Having read https://blogs.mathworks.com/cleve/2013/10/14/complex-step-differentiation/ thanks to a prompting by @tBuLi it seems to me that we might be able to get reasonable autodiff for many algorithms just using the built in complex number support of numpy (and numba). For example:
```

import numpy as np
import time
from clifford.g3c import *
from clifford.tools.g3c import fast_up
import numba

@numba.njit
def diff_up(x, dxda):
    """
    The derivative of the up function in clifford
    """
    return (dxda|x).value[0]*einf + dxda

def test_diff_up():
    # Make some data
    a = np.random.randn()
    b = np.random.randn()
    c = np.random.randn()

    # Set up a little step
    x = a*e1 + b*e2 + c*e3

    # Calculate a the exact derivative
    dxda = e1
    exactdiffup = diff_up(x, dxda)
    print(exactdiffup)

    # Do the complex step derivative
    h = 1E-8
    diff_in = x + 1j*h*e1
    res = fast_up(diff_in)
    comp_step_up = layout.MultiVector(value=res.value.imag/h)
    print(comp_step_up)

    print(np.sum((exactdiffup.value - comp_step_up.value)**2))

    start_time = time.time()
    for i in range(10000):
        diff_up(x, dxda)
    end_time = time.time()
    print(end_time - start_time)

    start_time = time.time()
    for i in range(10000):
        fast_up(diff_in)
    end_time = time.time()
    print(end_time - start_time)

if __name__ == '__main__':
    test_diff_up()
```

PR #355, opened 2020-09-10
Add tests for complex number type support in algebras

hugohadfield:
Adds tests to ensure that complex number type support behaves as one would expect and type info is not lost, starts addressing https://github.com/pygae/clifford/issues/354

### DIFF
--- a/clifford/_layout.py
+++ b/clifford/_layout.py
@@ -574,7 +574,7 @@ class Layout(object):
 
         @_numba_utils.njit
         def comp_func(Xval):
-            Yval = np.zeros(dims)
+            Yval = np.zeros(dims, dtype=Xval.dtype)
             for i, s in enumerate(signlist):
                 Yval[i] = Xval[dims-1-i]*s
             return Yval
--- a/clifford/_layout.py
+++ b/clifford/_layout.py
@@ -447,7 +447,7 @@ class Layout(object):
         else:
             # Equivalent to but faster than
             #   Iinv = self.pseudoScalar.inv().value
-            Iinv = np.zeros(self.gaDims)
+            Iinv = np.zeros(self.gaDims, dtype=int)
             II_scalar = self.gmt[-1, 0, -1]
             # set the pseudo-scalar part
             Iinv[-1] = 1 / II_scalar
--- a/clifford/_layout.py
+++ b/clifford/_layout.py
@@ -447,10 +447,14 @@ class Layout(object):
         else:
             # Equivalent to but faster than
             #   Iinv = self.pseudoScalar.inv().value
-            Iinv = np.zeros(self.gaDims, dtype=int)
             II_scalar = self.gmt[-1, 0, -1]
+            inv_II_scalar = 1 / II_scalar
+            if II_scalar == 1 or -1:
+                Iinv = np.zeros(self.gaDims, dtype=int)
+            else:
+                Iinv = np.zeros(self.gaDims, dtype=type(inv_II_scalar))
             # set the pseudo-scalar part
-            Iinv[-1] = 1 / II_scalar
+            Iinv[-1] = inv_II_scalar
 
             gmt_func = self.gmt_func
             @_numba_utils.njit
--- a/clifford/_layout.py
+++ b/clifford/_layout.py
@@ -449,7 +449,7 @@ class Layout(object):
             #   Iinv = self.pseudoScalar.inv().value
             II_scalar = self.gmt[-1, 0, -1]
             inv_II_scalar = 1 / II_scalar
-            if II_scalar == 1 or -1:
+            if II_scalar in (1, -1):
                 Iinv = np.zeros(self.gaDims, dtype=int)
             else:
                 Iinv = np.zeros(self.gaDims, dtype=type(inv_II_scalar))
--- a/clifford/_multivector.py
+++ b/clifford/_multivector.py
@@ -60,8 +60,8 @@ class MultiVector(object):
                     self.layout.gaDims)
 
     def __array__(self) -> 'cf.MVArray':
-        # we are a scalar, and the only appropriate dtype is an object array
-        return cf.MVArray([self])
+        # ensure that coercion gives our array subclass
+        return cf.array(self)
 
     def _checkOther(self, other, coerce=True) -> Tuple['MultiVector', bool]:
         """Ensure that the other argument has the same Layout or coerce value if
--- a/clifford/_multivector.py
+++ b/clifford/_multivector.py
@@ -118,11 +118,42 @@ class MultiVector(object):
         functions instead, as these work in degenerate metrics like PGA too,
         and are equivalent but faster in other metrics.
         """
-        return self.layout.MultiVector(value=self.layout.vee_func(self.value, other.value))
+        other, mv = self._checkOther(other, coerce=True)
+
+        if mv:
+            newValue = self.layout.vee_func(self.value, other.value)
+        else:
+            if isinstance(other, np.ndarray):
+                obj = self.__array__()
+                return obj&other
+
+            newValue = self.value&other
+
+
+        return self.layout.MultiVector(newValue)
+
+    def _rvee(self, other) -> 'MultiVector':
+        """Right-hand vee product, :math:`NM`"""
+
+        other, mv = self._checkOther(other, coerce=True)
+
+        if mv:
+            newValue = self.layout.vee_func(other.value, self.value)
+        else:
+            if isinstance(other, np.ndarray):
+                obj = self.__array__()
+                return other&obj
+            newValue = other&self.value
+
+        return self._newMV(newValue)
 
     def __and__(self, other) -> 'MultiVector':
         """ ``self & other``, an alias for :meth:`~MultiVector.vee` """
         return self.vee(other)
+
+    def __rand__(self, other) -> 'MultiVector':
+        """ ``other & other``, an alias for :meth:`~MultiVector._rvee` """
+        return self._rvee(other)
 
     def __mul__(self, other) -> 'MultiVector':
         """ ``self * other``, the geometric product :math:`MN` """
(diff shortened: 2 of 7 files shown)


hugohadfield:
My knowledge of pytest.fixture is not very good, @eric-wieser any suggestions to get these tests to run over the conformalised metric as well?

clifford/test/test_complex.py
```diff
@@ -0,0 +1,100 @@
...
+        res = (A + 1j*B).value
+        res2 = A.value + 1j*B.value
+        np.testing.assert_allclose(res, res2)
+
+    def test_subtraction(self, algebra):
+        A = algebra.randomMV()
+        B = algebra.randomMV()
+        res = (A - 1j*B).value
+        res2 = A.value - 1j*B.value
+        np.testing.assert_allclose(res, res2)
```
eric-wieser, line 57: I think for these two tests you should require exact equality

clifford/test/test_complex.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def test_ip(self, algebra):
+        A = algebra.randomMV()
+        B = algebra.randomMV()
+        C = algebra.randomMV()
+        D = algebra.randomMV()
+        res = ((A + 1j*B)|(C + 1j*D)).value
+        res2 = (A|C).value + 1j*(B|C).value + 1j*(A|D).value - (B|D).value 
+        np.testing.assert_allclose(res, res2)
```
eric-wieser, line 84: Parametrization can merge these:
```suggestion
    @pytest.mark.parametrize('p', [cf.operator.gp, cf.operator.op, cf.operator.ip])
    def test_prod(self, algebra, p):
        A = algebra.randomMV()
        B = algebra.randomMV()
        C = algebra.randomMV()
        D = algebra.randomMV()
        res = (p(A + 1j*B, C + 1j*D)).value
        res2 = p(A, C).value + 1j*p(B, C).value + 1j*p(A, D).value - p(B, D).value 
        np.testing.assert_allclose(res, res2)
```
(with appropriate `import clifford as cf` or similar)

clifford/test/test_complex.py
```diff
@@ -0,0 +1,100 @@
...
+        res = (~(A + 1j*B)).value
+        res2 = (~A).value + 1j*(~B).value
+        np.testing.assert_allclose(res, res2)
+
+    def test_grade_selection(self, algebra):
+        A = algebra.randomMV()
+        B = algebra.randomMV()
+        res = ((A + 1j*B)(2)).value
+        res2 = A(2).value + 1j*B(2).value
+        np.testing.assert_allclose(res, res2)
```
eric-wieser, line 98: Again, you should require exact equality here.

clifford/test/test_complex.py
```diff
@@ -0,0 +1,100 @@
...
+
+
```
eric-wieser, line 100:
```suggestion
```

clifford/test/test_complex.py
```diff
@@ -0,0 +1,86 @@
...
+@pytest.fixture(scope='module')
+def g4_1():
+    return Cl(4, 1)[0]
+
+
+@pytest.fixture(scope='module')
+def g3c():
+    return conformalize(Cl(3)[0])[0]
```
eric-wieser, line 37: Using both seems unnecessary, given you're not actually using any of the other methods.

clifford/test/test_complex.py
```diff
@@ -0,0 +1,81 @@
+
```
eric-wieser, line 1: A docstring here would be good - perhaps something like

> the tests below are based on an informed guess about the correct behavior, and we should not be afraid of changing them if the things we're testing for turn out to not be the conventional definitions

clifford/_multivector.py
```diff
@@ -496,10 +496,10 @@ def __str__(self) -> str:
             else:
                 if coeff < 0:
                     sep = seps[1]
-                    abs_coeff = -round(coeff, p)
+                    abs_coeff = -np.round(coeff, p)
                 else:
                     sep = seps[0]
-                    abs_coeff = round(coeff, p)
+                    abs_coeff = np.round(coeff, p)
```
eric-wieser, line 502: This seems risky - it causes a crash for `str(2**64 * e1)`.

hugohadfield: So I did this to fix the deprecation warning:
```
clifford/test/test_complex.py: 53 warnings
  C:\Users\hughad01\internship\clifford\clifford\_multivector.py:499: DeprecationWarning: The Python built-in `round` is deprecated for complex scalars, and will raise a `TypeError` in a future release. Use `np.round` or `scalar.round` instead.
    abs_coeff = -round(coeff, p)
```
Should we use the scalar.round option instead?

eric-wieser: That doesn't exist for the case I give above either

hugohadfield:
Travis failures appear to be for unrelated reasons, squash and merge?

eric-wieser:
3.5 CI failure is caused by https://github.com/davidhalter/parso/issues/149

eric-wieser:
Other failures indeed look unrelated, although I'm not sure when they crept in

clifford/_layout.py
```diff
@@ -574,7 +574,7 @@ def _gen_complement_func(self, omt):
 
         @_numba_utils.njit
         def comp_func(Xval):
-            Yval = np.zeros(dims)
+            Yval = np.zeros(dims, dtype=Xval.dtype)
```
eric-wieser, line 577: Can you add a test to `test_clifford::TestClifford.test_unary_op_preserves_dtype` for this?

clifford/_multivector.py
eric-wieser: Can we either back out this change, or add a `if np.issubdtype(self.value.dtype, np.number` around the use of `np.round`?

clifford/_layout.py
```diff
@@ -447,7 +447,7 @@ def dual_func(self):
         else:
             # Equivalent to but faster than
             #   Iinv = self.pseudoScalar.inv().value
-            Iinv = np.zeros(self.gaDims)
+            Iinv = np.zeros(self.gaDims, dtype=int)
```
hugohadfield, line 450: @eric-wieser do you think this is a safe thing to do? I was aiming for a type that could be upgraded to whatever the type of the multivector was

eric-wieser: Ideally you'd pick the type based on the type of `1 / II_scalar`

eric-wieser: Otherwise this will go wrong for metrics containing 2

clifford/test/test_clifford.py
```diff
@@ -386,6 +386,8 @@ def test_binary_op_preserves_dtype(self, dtype, func, g3):
         operator.pos,
         operator.neg,
         MultiVector.gradeInvol,
+        MultiVector.dual,
+        MultiVector.right_complement
```
eric-wieser, line 390:
```suggestion
        MultiVector.right_complement,
        MultiVector.left_complement,
```

clifford/_layout.py
```diff
@@ -447,10 +447,14 @@ def dual_func(self):
         else:
             # Equivalent to but faster than
             #   Iinv = self.pseudoScalar.inv().value
-            Iinv = np.zeros(self.gaDims)
             II_scalar = self.gmt[-1, 0, -1]
+            inv_II_scalar = 1 / II_scalar
+            if II_scalar == 1 or -1:
```
eric-wieser, line 452:
```suggestion
            if II_scalar in (1, -1):
```

eric-wieser: What you had was nonsense

clifford/test/test_clifford.py
```diff
@@ -386,6 +386,8 @@ def test_binary_op_preserves_dtype(self, dtype, func, g3):
         operator.pos,
```
eric-wieser, line 386: I think you should be able to add `operator.and_,  # vee product` on line 375 now too.

clifford/_multivector.py
```diff
@@ -118,12 +118,42 @@ def vee(self, other) -> 'MultiVector':
...
+        other, mv = self._checkOther(other, coerce=True)
+
+        if mv:
+            newValue = self.layout.vee_func(self.value, other.value)
+        else:
+            if isinstance(other, np.ndarray):
+                obj = self.__array__()
+                return obj&other
+
+            newValue = self.value&other
```
eric-wieser, line 130: This line is wrong, it performs bitwise and. This code-path is taken by `(5*e1) & 3`, and should give the same result as if `3` were first turned into a real multivector.

clifford/_multivector.py
```diff
@@ -118,12 +118,42 @@ def vee(self, other) -> 'MultiVector':
...
+
+        return self.layout.MultiVector(newValue)
+
+    def _rvee(self, other) -> 'MultiVector':
```
eric-wieser, line 134: Hmm, this is gross.

clifford/_multivector.py
```diff
@@ -118,12 +118,42 @@ def vee(self, other) -> 'MultiVector':
...
         """
-        return self.layout.MultiVector(value=self.layout.vee_func(self.value, other.value))
+        other, mv = self._checkOther(other, coerce=True)
+
+        if mv:
+            newValue = self.layout.vee_func(self.value, other.value)
+        else:
+            if isinstance(other, np.ndarray):
+                obj = self.__array__()
+                return obj&other
```
eric-wieser, line 128: (note this line is fine, because here `obj` is a MVarray)

clifford/_multivector.py
hugohadfield: Surely in that case 3 would have been coerced into a `MultiVector` during `checkother(3, coerce=True)`?

eric-wieser: I missed the coerce=True. I'm not certain this line is reachable - so I'd be inclined to raise `TypeError`.

It's pretty obvious that our coercion rules needs some work. I'll maybe revisit this in another PR.

eric-wieser (approved):

PR #357, opened 2020-09-11
Preserve dtypes in left_complement, right_complement, and dual

eric-wieser:
Split from #355


eric-wieser:
CI failures are the usual, this code was written by Hugo anyway, so I'm going to go ahead and merge

PR #360, opened 2020-09-16
Speed up imports

hugohadfield:
Speed up imports (especially of conformal algebras) by removing the function preloading and adding some minor caching where possible

### DIFF
--- a/clifford/_layout.py
+++ b/clifford/_layout.py
@@ -560,8 +560,9 @@ class Layout(object):
 
         # Since we're working with basis blades, we can use the table directly.
         # We only care about the pseudo-scalar part of the wedge.
+        omt_ps_part = omt[:, -1, :]
         for n in range(dims):
-            signlist[n] = (-1)**(omt[n, -1, dims-1-n] < 0.001)
+            signlist[n] = (-1)**(omt_ps_part[n, dims-1-n] < 0.001)
 
         @_numba_utils.njit
         def comp_func(Xval):
--- a/clifford/_layout.py
+++ b/clifford/_layout.py
@@ -353,18 +353,6 @@ class Layout(object):
                 "names list of length %i needs to be of length %i" %
                 (len(names), self.gaDims))
 
-        # preload these lazy properties. Not doing this would likely be faster.
-        self.gmt_func
-        self.imt_func
-        self.omt_func
-        self.lcmt_func
-        self.adjoint_func
-        self.left_complement_func
-        self.right_complement_func
-        self.dual_func
-        self.vee_func
-        self.inv_func
-
     @property
     def gradeList(self):
         return list(self._basis_blade_order.grades)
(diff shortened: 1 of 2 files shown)


clifford/_layout.py
```diff
@@ -572,9 +560,8 @@ def _gen_complement_func(self, omt):
 
         # Since we're working with basis blades, we can use the table directly.
         # We only care about the pseudo-scalar part of the wedge.
-        omt_ps_part = omt[:, -1, :]
         for n in range(dims):
-            signlist[n] = (-1)**(omt_ps_part[n, dims-1-n] < 0.001)
+            signlist[n] = (-1)**(omt[n, -1, dims-1-n] < 0.001)
```
eric-wieser, line 564: Did you profile this? was it actually faster?

hugohadfield: After change:
image
Before change:
image
Good call! Will revert now :)

clifford/_layout.py
```diff
@@ -86,7 +86,7 @@ def omt_check(grade_v, grade_i, grade_j):
     return grade_v == (grade_i + grade_j)
 
 
-@_numba_utils.njit
+@_numba_utils.njit(cache=True)
```
eric-wieser, line 89: Any reason for not caching `_numba_construct_gmt`?

clifford/_layout.py
```diff
@@ -353,18 +353,6 @@ def __init__(self, *args, **kw):
...
-        self.gmt_func
-        self.imt_func
-        self.omt_func
-        self.lcmt_func
-        self.adjoint_func
-        self.left_complement_func
-        self.right_complement_func
-        self.dual_func
-        self.vee_func
-        self.inv_func
```
eric-wieser, line 366: I worry this change will have made this benchmark meaningless:

https://github.com/pygae/clifford/blob/ce0bc74520451a05ab156bd07ce3db783b4f0839/clifford/test/test_algebra_initialisation.py#L14-L22

Can you add these lines to the benchmark?

eric-wieser (commented):
Note that numba/numba#6373 should allow us to cache `_numba_construct_graded_mt` too - but no need to wait for that for this PR.

hugohadfield:
To test the import speed up thanks to this caching we did the following test:
1. Install clifford in developer mode
2. Run the following script, `time_import.py`, just to get the general python caching going etc. Don't bother recording the result
```
import numba, numpy, time, sparse 
t = time.perf_counter()
import clifford.g3c
print(time.perf_counter() - t)
```
3. In the `clifford/clifford/__pycache__` directory delete all `.nbi` and `.nbc` files
4. Run  `time_import.py` again to get the effect of jitted but uncached clifford functions
5. Run  `time_import.py` again to get  the effect of jitted and cached clifford functions

Results on my home PC:
Uncached 1.8571729790000973
Cached 0.9225598940010968

We also repeated the experiment but changing it also test the time required to also multiply two mvs:
```
import numba, numpy, time, sparse
t = time.perf_counter()
from clifford.g3c import *
print(e1*e2)
print(time.perf_counter() - t)
```
Results 
Uncached 1.9949228559999028
Cached 1.0691611509992072

PR #361, opened 2020-09-16
Do not coerce 0d MVArrays into 1d arrays

eric-wieser:
If a user asks for a 0d array, they should get a 0d array.


hugohadfield (approved):
Seems sensible and relevant tests pass 👍

PR #362, opened 2020-09-16
Start conversion of njit to jit to allow object mode multivectors

hugohadfield:
Doing this will allow mvs in object mode which will, for example, allow us to mix in sympy symbols


hugohadfield:
I remember we discussed with the Numba devs that the jit fallback is being deprecated, not sure what that means exactly for this PR

Issue #363, opened 2020-09-24
Jitted functions using global multivectors cannot be cached


hugohadfield:
This code:
```python
import numba
from clifford.g3c import *

@numba.njit(cache=True)
def foo(x):
    return (x * e3).value

print(foo(1))
```
Produces:
```
test_cache.py:4: NumbaWarning: Cannot cache compiled function "foo" as it uses dynamic globals (such as ctypes pointers and large global arrays)
  @numba.njit(cache=True)
[0 0 0 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
```

eric-wieser:
It's the multivector constant for some reason:
```python
In [12]: from clifford.g3 import *

In [13]: import warnings; warnings.simplefilter('error')

In [14]: k = 1*e1 + 20*e2 + 300*e3

In [15]: @numba.njit(cache=True)
    ...: def foo(x):
    ...:     return (x * k).value
    ...:

In [16]: foo(1)
---------------------------------------------------------------------------
NumbaWarning                              Traceback (most recent call last)
<ipython-input-16-7d8d25071659> in <module>
----> 1 foo(1)

<snip>

NumbaWarning: Cannot cache compiled function "foo" as it uses dynamic globals (such as ctypes pointers and large global arrays)

In [17]: %debug
> c:\users\wiese\repos\ga\numba\numba\core\caching.py(425)check_cachable()
    423             msg = ('Cannot cache compiled function "%s" %s'
    424                    % (cres.fndesc.qualname.split('.')[-1], cannot_cache))
--> 425             warnings.warn_explicit(msg, NumbaWarning,
    426                                    self._locator._py_file, self._lineno)
    427             return False

ipdb> var = next(gv for gv in cres.library._final_module.global_variables if gv.name.startswith('numba.dynamic.globals'))
ipdb> addr = int(str(var).split(" ")[-3])
ipdb> import ctypes
ipdb> p list((ctypes.c_uint32 * 8).from_address(addr))
[0, 1, 20, 300, 0, 0, 0, 0]  # this is definitely `k`
```

eric-wieser:
Caused by these lines:

https://github.com/numba/numba/blob/master/numba/core/base.py#L1033-L1034

Our multivector numba type does not use `F` or `C` order, because that was marginally more work

[...]

PR #364, opened 2020-09-24
Use C-order arrays where possible in numba extension types

hugohadfield:
Hopefully solves https://github.com/pygae/clifford/issues/363, allowing multivector constants to be used in cached functions; as only C-order arrays are candidates for caching.

Previously only A-order arrays were supported.

### DIFF
--- a/clifford/numba/_layout.py
+++ b/clifford/numba/_layout.py
@@ -25,7 +25,6 @@ class LayoutType(types.Dummy):
         self.obj = layout
         # cache of multivector types for this layout
         self._c_cache = {}
-        self._f_cache = {}
         self._a_cache = {}
         layout_name = layout_short_name(layout)
         if layout_name is not None:
@@ -39,7 +38,6 @@ class LayoutType(types.Dummy):
         # contains a self-reference.
         d = self.__dict__.copy()
         d['_c_cache'] = {}
-        d['_f_cache'] = {}
         d['_a_cache'] = {}
         return d
 
--- a/clifford/numba/_layout.py
+++ b/clifford/numba/_layout.py
@@ -23,7 +23,9 @@ from .._multivector import MultiVector
 class LayoutType(types.Dummy):
     def __init__(self, layout):
         self.obj = layout
-        # cache of multivector types for this layout
+        # Caches of multivector types for this layout, in numba C and A order.
+        # Having two caches is faster than a cache keyed by a tuple of `(order, dt)`,
+        # and every millisecond counts in `MultiVector._numba_type_`.
         self._c_cache = {}
         self._a_cache = {}
         layout_name = layout_short_name(layout)
--- a/clifford/numba/_layout.py
+++ b/clifford/numba/_layout.py
@@ -24,7 +24,9 @@ class LayoutType(types.Dummy):
     def __init__(self, layout):
         self.obj = layout
         # cache of multivector types for this layout
-        self._cache = {}
+        self._c_cache = {}
+        self._f_cache = {}
+        self._a_cache = {}
         layout_name = layout_short_name(layout)
         if layout_name is not None:
             name = "LayoutType({})".format(layout_name)
@@ -36,7 +38,9 @@ class LayoutType(types.Dummy):
         # the cache causes issues with numba's pickle modifications, as it
         # contains a self-reference.
         d = self.__dict__.copy()
-        d['_cache'] = {}
+        d['_c_cache'] = {}
+        d['_f_cache'] = {}
+        d['_a_cache'] = {}
         return d
 
 
--- a/clifford/numba/_multivector.py
+++ b/clifford/numba/_multivector.py
@@ -14,13 +14,11 @@ try:
     import numba.np.numpy_support as _numpy_support
     from numba.core.imputils import impl_ret_borrowed, lower_constant
     from numba.core import cgutils, types
-    from numba.core.typing.typeof import typeof_impl
 except ImportError:
     # module locations prior to numba 0.49.0
     import numba.numpy_support as _numpy_support
     from numba.targets.imputils import impl_ret_borrowed, lower_constant
     from numba import cgutils, types
-    from numba.typing.typeof import typeof_impl
 
 from .._multivector import MultiVector
 
@@ -28,7 +26,6 @@ from ._layout import LayoutType
 from ._overload_call import overload_call
 
 __all__ = ['MultiVectorType']
-_typeof_ndarray = typeof_impl.dispatch(np.ndarray)
 
 
 class MultiVectorType(types.Type):
--- a/clifford/numba/_multivector.py
+++ b/clifford/numba/_multivector.py
@@ -34,7 +34,7 @@ _typeof_ndarray = typeof_impl.dispatch(np.ndarray)
 class MultiVectorType(types.Type):
     def __init__(self, layout: LayoutType, dtype: types.DType):
         self.layout_type = layout
-        self._array_type = dtype
+        self.value_type = dtype
         super().__init__(name='MultiVector({!r}, {!r})'.format(
             self.layout_type, self._array_type
         ))
@@ -42,10 +42,6 @@ class MultiVectorType(types.Type):
     @property
     def key(self):
         return self.layout_type, self._array_type
-
-    @property
-    def value_type(self):
-        return self._array_type
 
 
 # The docs say we should use register a function to determine the numba type
--- a/clifford/numba/_multivector.py
+++ b/clifford/numba/_multivector.py
@@ -54,6 +54,9 @@ def _numba_type_(self):
     cache = layout_type._cache
     dt = self.value.dtype
 
+    if not self.value.flags.c_contiguous:
+        raise ValueError('Clifford currently only supports JITing functions of MultiVectors with C_CONTIGUOUS value arrays.')
+
     # now use the dtype to key that cache.
     try:
         return cache[dt]
--- a/clifford/numba/_multivector.py
+++ b/clifford/numba/_multivector.py
@@ -56,16 +56,15 @@ def _numba_type_(self):
 
     cache = layout_type._cache
     dt = self.value.dtype
-    C_contig = self.value.flags.C_CONTIGUOUS
 
     # now use the dtype to key that cache.
     try:
-        return cache[(dt, C_contig)]
+        return cache[dt]
     except KeyError:
         # Computing and hashing `dtype_type` is slow, so we do not use it as a
         # hash key. The raw numpy dtype is much faster to use as a key.
         dtype_type = _typeof_ndarray(self.value, None)
-        ret = cache[(dt, C_contig)] = MultiVectorType(layout_type, dtype_type)
+        ret = cache[dt] = MultiVectorType(layout_type, dtype_type)
         return ret
 
 MultiVector._numba_type_ = _numba_type_
--- a/clifford/numba/_multivector.py
+++ b/clifford/numba/_multivector.py
@@ -51,20 +51,25 @@ class MultiVectorType(types.Type):
 def _numba_type_(self):
     layout_type = self.layout._numba_type_
 
-    cache = layout_type._cache
+    relevant_cache = layout_type._c_cache
     dt = self.value.dtype
 
     if not self.value.flags.c_contiguous:
-        raise ValueError("Only clifford.MultiVector objects with contiguous coefficients can be passed to JITted functions")
+        if self.value.flags.f_contiguous:
+            relevant_cache = layout_type._f_cache
+        elif self.value.flags.aligned:
+            relevant_cache = layout_type._a_cache
+        else:
+            raise ValueError("Not sure what to write here")
 
     # now use the dtype to key that cache.
     try:
-        return cache[dt]
+        return relevant_cache[dt]
     except KeyError:
         # Computing and hashing `value_type` is slow, so we do not use it as a
         # hash key. The raw numpy dtype is much faster to use as a key.
         value_type = _numpy_support.from_dtype(dt)[::1]  # c-order
-        ret = cache[dt] = MultiVectorType(layout_type, value_type)
+        ret = relevant_cache[dt] = MultiVectorType(layout_type, value_type)
         return ret
 
 MultiVector._numba_type_ = _numba_type_
--- a/clifford/numba/_multivector.py
+++ b/clifford/numba/_multivector.py
@@ -14,11 +14,13 @@ try:
     import numba.np.numpy_support as _numpy_support
     from numba.core.imputils import impl_ret_borrowed, lower_constant
     from numba.core import cgutils, types
+    from numba.core.typing.typeof import typeof_impl
 except ImportError:
     # module locations prior to numba 0.49.0
     import numba.numpy_support as _numpy_support
     from numba.targets.imputils import impl_ret_borrowed, lower_constant
     from numba import cgutils, types
+    from numba.typing.typeof import typeof_impl
 
 from .._multivector import MultiVector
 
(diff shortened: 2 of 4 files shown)


clifford/numba/_multivector.py
```diff
@@ -14,35 +14,38 @@
...
 
 __all__ = ['MultiVectorType']
+_typeof_ndarray = typeof_impl.dispatch(np.ndarray)
 
 
 class MultiVectorType(types.Type):
     def __init__(self, layout: LayoutType, dtype: types.DType):
         self.layout_type = layout
-        self._scalar_type = dtype
+        self._array_type = dtype
```
eric-wieser, line 37:
```suggestion
        self.value_type = dtype
```
and remove the `@property` below

clifford/numba/_multivector.py
```diff
@@ -64,7 +67,7 @@ def _numba_type_(self):
     except KeyError:
         # Computing and hashing `dtype_type` is slow, so we do not use it as a
         # hash key. The raw numpy dtype is much faster to use as a key.
-        dtype_type = _numpy_support.from_dtype(dt)
+        dtype_type = _typeof_ndarray(self.value, None)
         ret = cache[dt] = MultiVectorType(layout_type, dtype_type)
```
eric-wieser, line 71: This cache is now unsafe - you're caching both `int, C` and `int, F` under the same key, `int`.

clifford/test/test_function_cache.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+from numba import generated_jit
+import warnings
+
+# Make the test fail on a failed cache warning
+warnings.simplefilter("error")
```
eric-wieser, line 6: When put here, this will affect the whole test suite.

clifford/test/test_function_cache.py
```diff
@@ -0,0 +1,20 @@
...
+@generated_jit(cache=True)
+def foo(x):
+    from clifford.g3c import e3
+
+    def impl(x):
+        return (x * e3).value
+    return impl
+
+
+def test_function_cache():
```
eric-wieser, line 18:
```suggestion
@pytest.mark.filterwarnings("error")
def test_function_cache():
```

eric-wieser: Plus the comment you had before

clifford/test/test_function_cache.py
```diff
@@ -0,0 +1,20 @@
...
+from numba import generated_jit
+import warnings
+
+# Make the test fail on a failed cache warning
+warnings.simplefilter("error")
+
+
+@generated_jit(cache=True)
+def foo(x):
+    from clifford.g3c import e3
```
eric-wieser, line 11:
```suggestion
    from clifford.g3 import e3
```
May as well make this test faster :)

clifford/test/test_function_cache.py
hugohadfield: I was a bit worried about whether that might be the case

clifford/numba/_multivector.py
```diff
@@ -14,35 +14,34 @@
...
         ))
 
     @property
     def key(self):
-        return self.layout_type, self._scalar_type
-
-    @property
-    def value_type(self):
-        return self._scalar_type[:]
+        return self.layout_type, self._array_type
```
eric-wieser, line 44:
```suggestion
        return self.layout_type, self.value_type
```
throughout

clifford/numba/_multivector.py
hugohadfield: I guess we could cache with the whole `dtype_type` or maybe a tuple `(int, C)`?

eric-wieser: Caching with the whole `dtype_type` might be too slow, see the comment above.

clifford/numba/_multivector.py
```diff
@@ -57,15 +56,16 @@ def _numba_type_(self):
...
     try:
-        return cache[dt]
+        return cache[(dt, C_contig)]
     except KeyError:
         # Computing and hashing `dtype_type` is slow, so we do not use it as a
         # hash key. The raw numpy dtype is much faster to use as a key.
-        dtype_type = _numpy_support.from_dtype(dt)
-        ret = cache[dt] = MultiVectorType(layout_type, dtype_type)
+        dtype_type = _typeof_ndarray(self.value, None)
+        ret = cache[(dt, C_contig)] = MultiVectorType(layout_type, dtype_type)
```
eric-wieser, line 68: I'm starting to think that using `_typeof_ndarray` is a mistake after all - the only way to create a safe cache is to use it as the key, and that just costs way too much. For `a = np.ones(32)`, I measure:
```python
# the cost of a multivector addition
In [55]: %timeit a + a
1.16 µs ± 263 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
```python
# the current cache key
In [82]: d = {a.dtype: None}

In [83]: %timeit d[a.dtype]
208 ns ± 16.9 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [84]: %timeit d[a.dtype]
204 ns ± 36.5 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
```python
# your cache key
In [69]: d = {(a.dtype, a.flags.c_contiguous): None}

In [70]: %timeit d[(a.dtype, a.flags.c_contiguous)]
466 ns ± 129 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [71]: %timeit d[(a.dtype, a.flags.c_contiguous)]
491 ns ± 156 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
```python
# a cache key that includes the memory order
In [72]: d = {(a.dtype, map_layout(a)): None}

In [73]: %timeit d[(a.dtype, map_layout(a))]
716 ns ± 133 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [74]: %timeit d[(a.dtype, map_layout(a))]
824 ns ± 160 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
```python
# the full cache key
In [75]: d = {_typeof_ndarray(a, None): None}

In [76]: %timeit d[_typeof_ndarray(a, None)]
22.3 µs ± 926 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [77]: %timeit d[_typeof_ndarray(a, None)]
36.4 µs ± 7.76 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

I think our options are:

* Support _only_ C-order multivectors. Frankly, this is the most common case, and the cost of making a C-order copy is probably lower than the dispatch overhead. We don't need to change our caching if we do this. I don't know what the behavior is if you have a non-C order multivector. Ideally it silently copies, but it might refuse to compile.
* Support C-order and A-order multivectors (via your cache key). We should then construct the appropriate `Array` type ourselves instead of calling `_typeof_ndarray`, as that's the only way to ensure our cache is truthful. There's no point paying for F-order multivectors, as A is a superset of F that's just less efficient.

hugohadfield: Nice investigation, I think the C-only option makes sense (and is the most performant) as long as there are no specific really desirable use cases for A ordered arrays?

clifford/test/test_numba_extensions.py
```diff
@@ -174,15 +174,15 @@ def test_pickling():
...
 
     @numba.njit
-    def mul_mv(a):
-        return a*e3
+    def mul_mv(mv):
+        return mv*e3
 
-    mva = layout.MultiVector(np.ones(8, order='F'))
-    mvb = layout.MultiVector(np.ones(8, order='C'))
+    mva = layout.MultiVector(np.ones(8)[::-1])
```
eric-wieser, line 184:
```suggestion
    mva = layout.MultiVector(np.ones(8)[::-1])
    assert not mva.value.c_contiguous
```
to check this is actually testing what you want to test

clifford/numba/_multivector.py
```diff
@@ -64,7 +63,7 @@ def _numba_type_(self):
     except KeyError:
         # Computing and hashing `dtype_type` is slow, so we do not use it as a
         # hash key. The raw numpy dtype is much faster to use as a key.
-        dtype_type = _numpy_support.from_dtype(dt)
+        dtype_type = _typeof_ndarray(self.value, None)
```
eric-wieser, line 66:
```suggestion
        dtype_type = _numpy_support.from_dtype(dt)[::1]  # c-order
```

clifford/numba/_multivector.py
```diff
@@ -14,35 +14,34 @@
...
 
 __all__ = ['MultiVectorType']
+_typeof_ndarray = typeof_impl.dispatch(np.ndarray)
 
 
 class MultiVectorType(types.Type):
     def __init__(self, layout: LayoutType, dtype: types.DType):
         self.layout_type = layout
-        self._scalar_type = dtype
+        self.value_type = dtype
```
eric-wieser, line 37: Note that these changes are probably still a good idea - they make it possible to use `MultiVectorType` for any array order,, even if the type inference doesn't support it right now.

clifford/numba/_multivector.py
```diff
@@ -64,7 +63,7 @@ def _numba_type_(self):
     except KeyError:
         # Computing and hashing `dtype_type` is slow, so we do not use it as a
         # hash key. The raw numpy dtype is much faster to use as a key.
-        dtype_type = _numpy_support.from_dtype(dt)
+        dtype_type = _numpy_support.from_dtype(dt)[::1]  # c-order
         ret = cache[dt] = MultiVectorType(layout_type, dtype_type)
```
eric-wieser, line 67: These are now misnamed:
```suggestion
        value_type = _numpy_support.from_dtype(dt)[::1]  # c-order
        ret = cache[dt] = MultiVectorType(layout_type, value_type)
```

eric-wieser: And I suppose the comment above

clifford/test/test_function_cache.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+from numba import generated_jit
+import pytest
+
+
+@generated_jit(cache=True)
```
eric-wieser, line 6: This either need to use `_numba_utils`, or be disabled on runs when jitting is disabled

clifford/test/test_numba_extensions.py
eric-wieser: Also, `8` should be `layout.gaDims`, because you're in g3c here.

clifford/test/test_numba_extensions.py
```diff
@@ -172,3 +172,19 @@ def test_pickling():
...
+
+def test_A_order():
+    import numpy as np
+
+    @numba.njit
+    def mul_mv(mv):
+        return mv*e3
+
+    mva = layout.MultiVector(np.ones(8)[::-1])
+    assert not mva.value.c_contiguous
```
eric-wieser, line 185: This assert fails because `Mv.__init__` always copies the input.

You can fix this with something awkward like
```
mva = layout.MultiVector(np.ones(8))
mva.value = mva.value[::-1]
```

clifford/numba/_multivector.py
```diff
@@ -14,35 +14,34 @@
...
     from numba import cgutils, types
+    from numba.typing.typeof import typeof_impl
 
 from .._multivector import MultiVector
 
 from ._layout import LayoutType
 from ._overload_call import overload_call
 
 __all__ = ['MultiVectorType']
+_typeof_ndarray = typeof_impl.dispatch(np.ndarray)
```
eric-wieser, line 31: These changes are now redundant

hugohadfield:
So looks like the A order mvs do not behave nicely 
https://travis-ci.org/github/pygae/clifford/jobs/730198231#L414

eric-wieser:
Perhaps for now then we should raise an exception if `flags.c_contiguous` is not `True`? I don't think we care too much about allowing A-order arrays, but we certainly care about them producing wrong result silently.

clifford/numba/_multivector.py
```diff
@@ -54,6 +54,9 @@ def _numba_type_(self):
     cache = layout_type._cache
     dt = self.value.dtype
 
+    if not self.value.flags.c_contiguous:
+        raise ValueError('Clifford currently only supports JITing functions of MultiVectors with C_CONTIGUOUS value arrays.')
```
eric-wieser, line 58:
```suggestion
        raise ValueError("Only clifford.MultiVector objects with contiguous coefficients can be passed to JITted functions")
```
since for 1d arrays, all contiguous arrays are c-contiguous

clifford/test/test_numba_extensions.py
```diff
@@ -172,3 +172,20 @@ def test_pickling():
...
+        return mv*e3
+
+    mva = layout.MultiVector(np.ones(layout.gaDims))
+    mva.value = mva.value[::-1]
+    assert not mva.value.flags.c_contiguous
+
+    mvc = layout.MultiVector(np.ones(layout.gaDims))
+    resa = mul_mv(mva)
+    resc = mul_mv(mvc)
+    assert resa == resc
```
eric-wieser, line 191:
```suggestion
    mvc = layout.MultiVector(np.ones(layout.gaDims))
    
    with pytest.raises(ValueError):
        mul_mv(mva)
    
    # C-order is fine
    mul_mv(mvc)
```

clifford/test/test_numba_extensions.py
```diff
@@ -172,3 +172,23 @@ def test_pickling():
...
+        return mv*e3
+
+    mva = layout.MultiVector(np.ones(layout.gaDims))
+    mva.value = mva.value[::-1]
+    assert not mva.value.flags.c_contiguous
+
+    mvc = layout.MultiVector(np.ones(layout.gaDims))
+
+    with pytest.raises(ValueError):
+        mul_mv(mva)
```
eric-wieser, line 191: CI fails in two ways here:
* On jit-less mode, this doesn't error at all (as we'd expect) - we should skip this part of the test if jitting is disabled
* When jitting:
   * Numba catches this `ValueError`
   * Tries to recompile in object mode (`jit` not `njit`)
   * Fails, and throws `TypingError`

  The quick fix is to adjust the exception type - but falling back to object mode sounds like a dangerous thing to do performance-wise.

clifford/numba/_multivector.py
```diff
@@ -55,17 +51,25 @@ def value_type(self):
...
+    relevant_cache = layout_type._c_cache
     dt = self.value.dtype
 
+    if not self.value.flags.c_contiguous:
+        if self.value.flags.f_contiguous:
+            relevant_cache = layout_type._f_cache
+        elif self.value.flags.aligned:
+            relevant_cache = layout_type._a_cache
+        else:
+            raise ValueError("Not sure what to write here")
```
eric-wieser, line 63: Does `a` mean `aligned`? I thought it meant "any".

clifford/numba/_multivector.py
```diff
@@ -55,17 +51,25 @@ def value_type(self):
...
     try:
-        return cache[dt]
+        return relevant_cache[dt]
     except KeyError:
-        # Computing and hashing `dtype_type` is slow, so we do not use it as a
+        # Computing and hashing `value_type` is slow, so we do not use it as a
         # hash key. The raw numpy dtype is much faster to use as a key.
-        dtype_type = _numpy_support.from_dtype(dt)
-        ret = cache[dt] = MultiVectorType(layout_type, dtype_type)
+        value_type = _numpy_support.from_dtype(dt)[:]
```
eric-wieser, line 71: This `[:]` needs to be adjusted to match the order - `[::1]` is C, I forget what `F` is, and `[:]` is any. You could use `numba.types.Array` directly instead I think.

eric-wieser: https://github.com/numba/numba/blob/3434eac6683a915243d8c4c89d483f9a04fabca9/numba/np/numpy_support.py#L215-L222

clifford/numba/_multivector.py
```diff
@@ -53,19 +49,29 @@ def value_type(self):
...
+        return None
+
     layout_type = self.layout._numba_type_
 
-    cache = layout_type._cache
+    relevant_cache = layout_type._c_cache
     dt = self.value.dtype
 
+    if not self.value.flags.c_contiguous:
+        relevant_cache = layout_type._a_cache
```
eric-wieser, line 62: I'd have a slightly easier time reading this as:
```suggestion
    dt = self.value.dtype
    if self.value.flags.c_contiguous:
        relevant_cache = layout_type._c_cache
    else:
        relevant_cache = layout_type._a_cache
```

clifford/numba/_layout.py
```diff
@@ -24,7 +24,8 @@ class LayoutType(types.Dummy):
     def __init__(self, layout):
         self.obj = layout
         # cache of multivector types for this layout
```
eric-wieser, line 26:
```suggestion
        # Caches of multivector types for this layout, in numba C and A order.
        # Having two caches is faster than a cache keyed by a tuple of `(order, dt)`,
        # and every millisecond counts in `MultiVector._numba_type_`.
```

eric-wieser (approved):

PR #365, opened 2020-10-04
Add 2D PGA (Cl(2,0,1))

kliment:
Adds a 2D version of the projective GA using Cl(2,0,1) - this is wonderful for working with lines, points, areas, angles and distances in the plane.

### DIFF
--- a/clifford/pga2d.py
+++ b/clifford/pga2d.py
@@ -1,0 +1,8 @@
+from . import Cl
+
+layout, blades = Cl(2, 0, 1, firstIdx=0)
+locals().update(blades)
+
+# for shorter reprs
+layout.__name__ = 'layout'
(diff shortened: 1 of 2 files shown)


eric-wieser:
Could you add an entry to https://github.com/pygae/clifford/blob/master/docs/predefined-algebras.rst too?

kliment:
Done

kliment:
Just checking, do you need anything more from my end to merge this?

arsenovic:
@eric-wieser  what do you think? not sure about the tests, but seems like a pretty simple addition?

eric-wieser:
There was a discussion on our slack about what to name the algebra, we should check if that came to a conclusion before merging.

Issue #366, opened 2020-10-08
Pseudoinverse for multivectors


tBuLi:
Would it be possible to add an option to use the matrix pseudoinverse instead of the normal matrix inverse into the method for computing inverses for multivectors? Having read the relevant chapter in Christian Perwass' book, he also mentions there that a pseudoinverse should work. I think it would be very nice to add an option to e.g. the `.inv()` method that allows something like

```python
M.inv(pseudo=True)
```
given that in some situations this is still meaningful to do.

eric-wieser:
I assume this is for degenerate basis vectors, when sometimes there is no inverse?

tBuLi:
Yes that's one case, and another one I've found is when performing bivector splits when the two bivectors in the split B = B_1 + B_2 start to become to similar. When they become identical the real inverse gives a LinAlgError, but the pseudoinverse returns B_1 = B_2 = B / 2. This is much friendlier ;).

eric-wieser:
If you update to the development version of `clifford`, you should find that the linear algebra method is not used at all.

hugohadfield:
I think we still use it for algebras greater than dimension 6?

tBuLi:
I've taking the liberty of making a PR for this with a slightly different API then suggested above: I added a `.pinv()` method to really make it easy to distinguish. Let me know if you like it or want to see some changes!

PR #367, opened 2020-10-09
Added a .pinv() method to MultiVectors plus tests

tBuLi:
PR implementing #366. 

I've added a method `.pinv()` to multivectors, which returns the pseudoinverse instead of the normal inverse. Identical to the matrix inverse method but when invoked on singular elements this does not throw an error but instead returns the original element, because well... it's singular.

### DIFF
--- a/clifford/_layout.py
+++ b/clifford/_layout.py
@@ -690,6 +690,29 @@ class Layout(object):
 
         return leftLaInvJIT
 
+    @_cached_property
+    def pinv_func(self):
+        """
+        Get a function that returns left-pseudoinverse using a computational
+        linear algebra method proposed by Christian Perwass. Identical to
+        `inv_func` except the inverse has been replaced by a pseudo inverse.
+        """
+        mult_table = self.gmt
+        k_list, l_list, m_list = mult_table.coords
+        mult_table_vals = mult_table.data
+        n_dims = mult_table.shape[1]
+
+        identity = np.zeros((n_dims,))
+        identity[self._basis_blade_order.bitmap_to_index[0]] = 1
+
+        @_numba_utils.njit
+        def leftLaPInvJIT(value):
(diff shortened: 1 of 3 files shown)


clifford/_layout.py
```diff
@@ -690,6 +690,29 @@ def leftLaInvJIT(value):
 
         return leftLaInvJIT
 
+    @_cached_property
+    def pinv_func(self):
+        """
+        Get a function that returns left-pseudoinverse using a computational
```
eric-wieser, line 696: Is the left pseudo-inverse equal to the right pseudo-inverse (i.e., can you find a counterexample where it isn't?)

eric-wieser:
CI failures can be ignored. Thanks for the PR, I think I like the `pinv` name better than your previous suggestion anyway.

clifford/_multivector.py
```diff
@@ -773,6 +773,9 @@ def inv(self) -> 'MultiVector':
     leftInv = leftLaInv
     rightInv = leftLaInv
 
+    def pinv(self) -> 'MultiVector':
+        return self._newMV(self.layout.pinv_func(self.value))
```
eric-wieser, line 777: Mind moving the docstring to here instead, since this is likely where it will be seen?

clifford/_layout.py
tBuLi: Good question. I can't come up with a counter example. Something interested and perhaps related I noticed is that if I replace the original np.linalg.solve with np.linalg.lstsq I get for he test example
```
Expected :0.5 + (3002399751580329.5^e2) - (0.5^e3) - (0.75^e13) - (3002399751580329.5^e23)
Actual   :0.5 - (0.5^e3)
```
I think it is save to say that this is undesired behavior. The pinv function behaves much gentler. And in the numpy docs I see no mention of left and right actually, so perhaps we should leave that distinction altogether? Since it seems to return the input element anyway in case of singular behavior I don't see how that could be different left and right.

PR #371, opened 2020-10-21
Add general trig functions via taylor series

hugohadfield:


clifford/__init__.py
```diff
@@ -258,17 +260,73 @@ def general_exp(x, max_order=15):
     for i in range(1, max_order):
         if np.any(np.abs(tmp.value) > _settings._eps):
             tmp = tmp*scaled * (1.0 / i)
-            result += tmp
+            result = result + tmp
```
eric-wieser, line 263: What's the motivation for this change?

clifford/test/test_trig_functions.py
```diff
@@ -0,0 +1,79 @@
...
+import numpy as np
+import pytest
+
+
+class TestScalarProperties:
+
+    @pytest.fixture()
+    def element(self):
+        alg, blades = Cl(0, 0, 0)
+        return blades['']
```
eric-wieser, line 12: alg.scalar might be clearer

clifford/__init__.py
hugohadfield: Motivation was that without this numba will not JIT our general_exp function:
```
No implementation of function Function(<built-in function iadd>) found for signature:
iadd(MultiVector(LayoutType(Layout([1, 1, 1], ids=BasisVectorIds.ordered_integers(3), order=BasisBladeOrder.shortlex(3), names=['', 'e1', 'e2', 'e3', 'e12', 'e13', 'e23', 'e123'])), float64), MultiVector(LayoutType(Layout([1, 1, 1], ids=BasisVectorIds.ordered_integers(3), order=BasisBladeOrder.shortlex(3), names=['', 'e1', 'e2', 'e3', 'e12', 'e13', 'e23', 'e123'])), float64))
```

clifford/test/test_trig_functions.py
```diff
@@ -0,0 +1,79 @@
+
+from .. import Cl
+import numpy as np
+import pytest
+
+
+class TestScalarProperties:
```
eric-wieser, line 7: Can you test this for complex numbers too using `cmath`?

eric-wieser: (you can convert from Multivectors in that algebra to complex with something like `mv.value.astype(float).view(complex)[0]`

clifford/_layout.py
```diff
@@ -589,11 +589,11 @@ def _hitzer_inverse(self):
         """
         Performs the inversion operation as described in the paper :cite:`Hitzer_Sangwine_2017`
         """
-        tot = np.sum(self.sig != 0)
+        tot = len(self.sig)
```
hugohadfield, line 592: Note this is needed to get the correct behaviour for the tan test, presumably this is the solution for degenerate metric algebras but it would be nice to see a proof of this

clifford/_layout.py
```diff
@@ -589,11 +589,11 @@ def _hitzer_inverse(self):
...
         @_numba_utils.njit
         def hitzer_inverse(operand):
             if tot == 0:
-                numerator = operand.layout.scalar
+                numerator = 1 + 0*operand
```
eric-wieser, line 596: What motivates this change?

hugohadfield: Currently LayoutType does not provide `.scalar` so the JIT fails unless you do something like this

eric-wieser: How was this never a problem before? I guess the code path got optimized out!

clifford/test/test_trig_functions.py
```diff
@@ -0,0 +1,87 @@
...
+    def test_trig_Cl010(self, Cl010element, np_func, cmath_func):
+        """
+        This tests the a clifford algebra isomorphic to the complex numbers
+        """
+        for x in np.linspace(0, 2 * np.pi, 10):
+            for y in np.linspace(0, 2 * np.pi, 10):
+                complex_mv = x * Cl010element[0] + y * Cl010element[1]
+                complex_value = x + 1j * y
+                result = np_func(complex_mv)
+                assert abs(result.value[0] + 1j * result.value[1] - cmath_func(complex_value)) < 1E-10
```
eric-wieser, line 70: I sent you down the rabbit hole with `cmath_func` here - `np.cos` works on complex numbers just fine - I had falsely assumed your tests were using `math.cos`.

clifford/_layout.py
hugohadfield: I think before we never tested it with Cl(0,0,0) it never hit a case where `tot == 0`!

clifford/test/test_trig_functions.py
hugohadfield: No worries, I'll remove the cmath import and we can just stick to numpy then :)

clifford/__init__.py
```diff
@@ -258,17 +260,77 @@ def general_exp(x, max_order=15):
...
+        op = op + (1 / math.gamma(2 * n + 1)) * X ** (2 * n)
+    return op
+
+
+def general_tanh(X, max_order=30):
+    """
+    The tanh function as the ratio of sinh and cosh
+    Note. It would probably be better to implement this as its own taylor series.
+    """
+    return general_sinh(X, max_order) / general_cosh(X, max_order)
```
eric-wieser, line 331: Should we put these in a new `taylor_expansions` submodule?

hugohadfield: I think that is probably a good idea

clifford/taylor_expansions.py
```diff
@@ -0,0 +1,117 @@
+
```
eric-wieser, line 1: Any chance you could add a docstring here, and add `taylor_expansions` to an rst file so that docs generate?

clifford/__init__.py
```diff
@@ -232,43 +232,6 @@ def grade_obj_func(objin_val, gradeList, threshold):
     return np.argmax(modal_value_count)
 
 
-def general_exp(x, max_order=15):
```
eric-wieser, line 235: Strictly you now need to do `from .taylor_expansions import general_exp` to make this still publicly accessible.

clifford/taylor_expansions.py
```diff
@@ -0,0 +1,117 @@
...
+import math
+
+import numpy as np
+
+from . import _numba_utils
+from . import _settings
+
+
+@_numba_utils.njit
+def general_exp(x, max_order=15):
```
eric-wieser, line 11: I'd argue you can now remove all the `general_` prefixes

clifford/__init__.py
eric-wieser: Or ``from .taylor_expansions import exp as general_exp` with my suggestion below. Or even:
```python
def general_exp(x, **kwargs):
   warnings.warn("cf.general_exp is deprecated. Use `mv.exp()` or `np.exp(mv)` on multivectors, or `cf.taylor_expansions.exp(x)` on arbitrary objects", DeprecationWarning)
   return taylor_expansions.exp(x, **kwargs)
```

clifford/_multivector.py
```diff
@@ -6,11 +6,16 @@
...
 
 import clifford as cf
-from .taylor_expansions import general_exp, general_sin, general_cos, general_tan, general_sinh, general_cosh, general_tanh
+import clifford.taylor_expansions as taylor_expansions
 from . import _settings
 from ._layout_helpers import layout_short_name
 
 
+def general_exp(x, **kwargs):
+   warnings.warn("cf.general_exp is deprecated. Use `mv.exp()` or `np.exp(mv)` on multivectors, or `cf.taylor_expansions.exp(x)` on arbitrary objects", DeprecationWarning)
```
eric-wieser, line 15:
```suggestion
   warnings.warn("cf.general_exp is deprecated. Use `mv.exp()` or `np.exp(mv)` on multivectors, or `cf.taylor_expansions.exp(x)` on arbitrary objects", DeprecationWarning, stacklevel=2)
```

clifford/numba/_multivector.py
```diff
@@ -250,6 +250,20 @@ def impl(a, b):
         return impl
 
 
+@numba.extending.overload(operator.pow)
+def ga_pow(a, b):
+    if isinstance(a, MultiVectorType) and isinstance(b, types.Integer):
+        gmt_func = a.layout_type.obj.gmt_func
+        def impl(a, b):
+            if b == 0:
```
eric-wieser, line 258: What about when `b < 0`? For now, `raise NotImplementedError` is probably fine.

Looks like `MultiVector.__pow__` is broken too.

eric-wieser:
This will need a merge / rebase in order to pull in the doc build pinning

hugohadfield:
I rebased and force pushed, hopefully this now works

docs/api/index.rst
```diff
@@ -10,3 +10,4 @@ API
     tools
     operator
     transformations
+    taylor_expansions
```
eric-wieser, line 13: This doesn't seem to be working - perhaps there's a warning in the sphinx log?

clifford/_layout.py
eric-wieser: Leaving this conversation unresolved so I remember to open an issue about making this less ugly.

clifford/_multivector.py
```diff
@@ -99,7 +99,25 @@ def _newMV(self, newValue=None, *, dtype: np.dtype = None) -> 'MultiVector':
     # binary
 
     def exp(self) -> 'MultiVector':
```
eric-wieser, line 101: May as well copy this annotation to all the methods below it

clifford/taylor_expansions.py
```diff
@@ -0,0 +1,140 @@
...
+    op = 1 + 0*X
+    X2 = X * X
+    X2n = 1 + 0*X
+    for n in range(1, max_order):
+        X2n = X2n*X2
+        op = op + ((-1) ** (n) / math.gamma(2 * n + 1)) * X2n
+    return op
+
+
+def tan(X, max_order=30):
```
eric-wieser, line 99: You should either `register_jittable` (or whatever it's called) or just use `njit` again here

hugohadfield: Currently I think we don't implement an overloaded multivector truediv in numba/_multivector so can't JIT this I think

eric-wieser: Good point. Maybe just a comment about that here and in tanh then, either in the docstring or a regular comment

hugohadfield: Added the comments. I guess sorting out the shirokov inverse PR will be my next job and then perhaps we can implement truediv calling that :)

eric-wieser:
I think my last comment is wondering whether we can make these look more like general_exp, and whether that improves the numerical stability. For instance, we could replace `gamma (2*n+1)` with iterated division by `2*n*(2*n+1)` or whatever the not-off-by-one version is.

We can always use that for a follow-up PR though, now that this PR adds tests

hugohadfield:
> 
> 
> I think my last comment is wondering whether we can make these look more like general_exp, and whether that improves the numerical stability. For instance, we could replace `gamma (2*n+1)` with iterated division by `2*n*(2*n+1)` or whatever the not-off-by-one version is.
Yes I think this would be a good idea to look at, I am sure that there will be some useful gains

>
> 
> We can always use that for a follow-up PR though, now that this PR adds tests
This is my thoughts too

clifford/_layout.py
eric-wieser: #385

clifford/taylor_expansions.py
```diff
@@ -0,0 +1,142 @@
...
+
+=================================================
+taylor_expansions (:mod:`clifford.taylor_expansions`)
+=================================================
+
+.. versionadded:: 1.3.2
+
+This file implements various Taylor expansions for useful functions of multivectors.
+For some algebra signatures there may exist closed forms of these functions which would likely be faster
+and more accurate. Nonetheless having pre-written taylor expansions for the general case is useful.
```
eric-wieser, line 12:
```suggestion
and more accurate. Nonetheless, having pre-written taylor expansions for the general case is useful.

.. note::
    Many of these functions are also exposed as :class:`~clifford.MultiVector` methods,
    such as :meth:`clifford.MultiVector.sin`. This means that ``mv.sin()`` or even ``np.sin(mv)`` can be used
    as a convenient interface to functions in this module, without having to import it directly.
    
    For example::
    
    >>> from clifford.g3 import *
    >>> import numpy as np
    >>> np.sin(np.pi*e12/4)
    # what does this give?
```

clifford/taylor_expansions.py
```diff
@@ -0,0 +1,142 @@
...
+import numpy as np
+
+from . import _numba_utils
+from . import _settings
+
+
+@_numba_utils.njit
+def exp(x, max_order=15):
+    """
+    This implements the series expansion of e**mv where mv is a multivector
```
eric-wieser, line 36:
```suggestion
    This implements the series expansion of :math:`\exp x` where :math:`x` is a multivector
```

clifford/taylor_expansions.py
```diff
@@ -0,0 +1,142 @@
...
+    The parameter order is the maximum order of the taylor series to use
```
eric-wieser, line 37:
```suggestion
    The parameter `max_order` is the maximum order of the taylor series to use
```

clifford/taylor_expansions.py
```diff
@@ -0,0 +1,142 @@
...
+        X2n = X2n*X2
+        op = op + ((-1) ** (n) / math.gamma(2 * n + 1)) * X2n
+    return op
+
+
+def tan(X, max_order=30):
+    """
+    The tan function as the ratio of sin and cos
+    Note. It would probably be better to implement this as its own taylor series. This function
+    is not JITed as currently we do not overload the truediv operator for multivectors.
```
eric-wieser, line 103:
```suggestion

    .. note::
        It would probably be better to implement this as its own taylor series. This function
        is not JITed as currently we do not overload the truediv operator for multivectors.
```

clifford/taylor_expansions.py
```diff
@@ -0,0 +1,142 @@
...
+        X2n = X2n * X2
+        op = op + (1 / math.gamma(2 * n + 1)) * X2n
+    return op
+
+
+def tanh(X, max_order=30):
+    """
+    The tanh function as the ratio of sinh and cosh
+    Note. It would probably be better to implement this as its own taylor series. This function
+    is not JITed as currently we do not overload the truediv operator for multivectors.
```
eric-wieser, line 140:
```suggestion

    .. note::
        It would probably be better to implement this as its own taylor series. This function
        is not JITed as currently we do not overload the truediv operator for multivectors.
```

clifford/test/test_trig_functions.py
```diff
@@ -0,0 +1,86 @@
+
+from .. import Cl
```
eric-wieser, line 2: Relative imports in tests are usually a bad idea - tests shouldn't assume they are part of the package.

```suggestion
from clifford import Cl
```

clifford/test/test_trig_functions.py
```diff
@@ -0,0 +1,86 @@
...
+        return alg.scalar, blades['e1']
+
+    @pytest.mark.parametrize('func, deriv_func', [(np.sin, np.cos),
+                                                  (np.cos, lambda x: -np.sin(x)),
+                                                  (np.tan, lambda x: (1/np.cos(x)**2)),
+                                                  (np.sinh, np.cosh),
+                                                  (np.cosh, np.sinh),
+                                                  (np.tanh, lambda x: (1 - np.tanh(x)**2))])
+    def test_derivatives(self, element, func, deriv_func):
+        for x in np.linspace(0, 2*np.pi, 10):
```
eric-wieser, line 37: How do these fare outside `[0, 2*pi]`? I'd be inclined to expand this to at least include some negative values,
```suggestion
        for x in np.linspace(-2*np.pi, 2*np.pi, 13):
```

eric-wieser (commented):
Earlier review was on mobile, apologies. A few small ends to tie up:

* I've made some doc suggestions above to try and steer users towards how to use this, and to tweak some formatting.
* This needs a release note, and it's easier to add it now than to add it in later
* The tests should cover some negative coefficients too!

clifford/taylor_expansions.py
```diff
@@ -5,11 +5,23 @@
...
+    Many of these functions are also exposed as :class:`~clifford.MultiVector` methods,
+    such as :meth:`clifford.MultiVector.sin`. This means that ``mv.sin()`` or even ``np.sin(mv)`` can be used
+    as a convenient interface to functions in this module, without having to import it directly.
+
+    For example::
+
+    >>> from clifford.g3 import *
+    >>> import numpy as np
+    >>> np.sin(np.pi*e12/4)
+    >>> # Produces (0.86867^e12)
```
eric-wieser, line 24:
```suggestion
    >>> np.sin(np.pi*e12/4)
    (0.86867^e12)
```

docs/changelog.rst
```diff
@@ -12,6 +12,10 @@ Changes in 1.4.x
  * Where possible, ``MultiVector``\ s preserve their data type in the dual, and
    the right and left complements.
 
+ * A new :mod:`clifford.taylor_expansions` is added to implement taylor series of various
+   multivector function, starting with common trigonometric functions. These functions are
+   additionally exposed via a
```
eric-wieser, line 17:
```suggestion
   additionally exposed via an incomplete sentence that Hugo forgot to write.
```

clifford/taylor_expansions.py
```diff
@@ -0,0 +1,164 @@
+"""
+.. currentmodule:: clifford.taylor_expansions
+
+=================================================
+taylor_expansions (:mod:`clifford.taylor_expansions`)
+=================================================
+
+.. versionadded:: 1.4.1
```
eric-wieser, line 8:
```suggestion
.. versionadded:: 1.4.0
```

clifford/taylor_expansions.py
```diff
@@ -0,0 +1,164 @@
+"""
+.. currentmodule:: clifford.taylor_expansions
+
+=================================================
+taylor_expansions (:mod:`clifford.taylor_expansions`)
+=================================================
```
eric-wieser, line 6:
```suggestion
=====================================================
taylor_expansions (:mod:`clifford.taylor_expansions`)
=====================================================
```

eric-wieser:
Thanks @hugohadfield :tada:

PR #372, opened 2020-10-22
fix radius calculation

### LINKED ISSUE #184: cga.Round.radius is always 1
```python
>>> from clifford.cga import CGA
>>> cga = CGA(3)
>>> r = cga.round()
>>> r.radius
1
```

pavlo-melnyk:
Closes #184

### DIFF
--- a/clifford/cga.py
+++ b/clifford/cga.py
@@ -302,7 +302,7 @@ class Round(CGAThing):
         radius of the round (a float)
         '''
         dual_sphere = self.dual
-        dual_sphere / (-dual_sphere | self.cga.einf)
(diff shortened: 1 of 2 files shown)


eric-wieser:
There are some tests in `test_cga.py` that are marked xfail because of this bug. Can you remove the marker?

pavlo-melnyk:
Please clarify which marker

eric-wieser:
Markers that mention the issue number you're fixing, like

https://github.com/pygae/clifford/blob/67ea8ae7b7b8004c4c8322949918d96b00e87fa7/clifford/test/test_cga.py#L84

eric-wieser:
Thanks!

PR #374, opened 2020-10-29
Documentation: Fix/add links to installation instructions

corcoted:
I noticed there were no links to installation instructions on the published versions of
* `README.md`
* `docs/index.rst`
So, I added simple instructions to `README.md` plus a link to the documentation page, and fixed the broken link in the documentation index.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,20 @@ Scalars, vectors, and higher-grade entities can be mixed freely and consistently
 ![Visual explanation of blades](https://raw.githubusercontent.com/pygae/clifford/master/docs/_static/blades.png)
 
 
+Quick Installation
+------------
+Requires Python version >=3.7
+
+Install using `conda`:
+```
+conda install clifford -c conda-forge
+```
+Install using `pip`:
+```
+pip3 install clifford
+```
+[Detailed instructions](https://slides.com/hugohadfield/installing-python-and-clifford)
+
 Quickstart
 ----------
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -25,6 +25,20 @@ Scalars, vectors, and higher-grade entities can be mixed freely and consistently
 ![Visual explanation of blades](https://raw.githubusercontent.com/pygae/clifford/master/docs/_static/blades.png)
 
 
+Quick Installation
+------------
+Requires Python version >=3.7
```
eric-wieser, line 30: I think we support 3.5 still?

corcoted: I was following what @hugohadfield wrote here: https://slides.com/hugohadfield/installing-python-and-clifford#/1 , but maybe that was just the latest python at the time.

hugohadfield: Ah yeah, I think those slides were for an old seminar we did and I probably just added the latest version of python at the time... Our CI runs (and often passes) on 3.5+ :)

hugohadfield: Also, thanks for the PR Ted! With the change to `version >=3.5` we should be OK to merge, what do you think @eric-wieser ?

corcoted:
Ok.  Fixed the python version in the "Quick Installation" section of README.md.  Should be good to go.

README.md
```diff
@@ -25,6 +25,20 @@ Scalars, vectors, and higher-grade entities can be mixed freely and consistently
 ![Visual explanation of blades](https://raw.githubusercontent.com/pygae/clifford/master/docs/_static/blades.png)
 
 
+Quick Installation
+------------
```
eric-wieser, line 29:
```suggestion
Quick Installation
------------------
```

PR #382, opened 2020-12-27
Fix the docbuilds with sphinxcontrib-bibtex>=2.0

eric-wieser:
The new version of sphinxcontrib-bibtex shows citation information in a tooltip:

image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,6 +58,8 @@ extensions = [
     #'sphinx.ext.mathjax',
 ]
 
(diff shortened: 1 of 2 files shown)


hugohadfield:
So this docs build was successful, and we don't need to pin the dependancy less than 2.0.0?

eric-wieser:
I think the resulting docs may have broken sphinx links though; but maybe only the new features in 2.0 are broken

PR #386, opened 2020-12-27
Fix a CI failure on Python 3.5

eric-wieser:
Parso is only used by our tests, as a transitive dependency of IPython

Here's a failing run: https://travis-ci.org/github/pygae/clifford/jobs/751695858#L2902


eric-wieser:
Looks like this did the trick - the failure is now a run-of-the-mill numerical issue rather than a CI failure, so I'm going to merge this so other PRs have a chance at getting a green tick again.

PR #389, opened 2020-12-28
updated (and broke some) tools,  add some docs. 

arsenovic:
updated some frame2mat, mat2frame methods to take a psuedo-scalar for frame instead of a layout, which is unneccsary and restrictive. 
also updated some docs.

### DIFF
--- a/clifford/test/test_tools.py
+++ b/clifford/test/test_tools.py
@@ -9,15 +9,17 @@ from clifford import Cl
 from clifford.tools import orthoFrames2Versor as of2v
 from clifford._numba_utils import DISABLE_JIT
 
+from clifford import tools
+
 
 too_slow_without_jit = pytest.mark.skipif(
     DISABLE_JIT, reason="test is too slow without JIT"
 )
 
 
-@unittest.skip("reason unknown")
+#@unittest.skip("reason unknown")
 class ToolsTests(unittest.TestCase):
-
+    @unittest.skip("reason unknown") 
     def checkit(self, p, q):
         # p, q =4,0
         N = p + q
@@ -39,7 +41,8 @@ class ToolsTests(unittest.TestCase):
 
         # Determined Versor implements desired transformation
         self.assertTrue([R_found*a*~R_found for a in A] == B)
-
+    
+    @unittest.skip("reason unknown") 
     def testOrthoFrames2VersorEuclidean(self):
         for p, q in [(2, 0), (3, 0), (4, 0)]:
             self.checkit(p=p, q=q)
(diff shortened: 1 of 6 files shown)


docs/tutorials/g3-algebra-of-space.ipynb
```diff
@@ -44,6 +44,7 @@
    "outputs": [],
    "source": [
     "import clifford as cf\n",
+    "from numpy import pi,e\n",
```
eric-wieser, line 47: I think this isn't needed - we fixed this problem in a different way, but because you're using your `master` branch to make PRs this is including the changes from your previous PR

eric-wieser:
```suggestion
```

arsenovic: what branch should i use to make pr?

eric-wieser: A new branch for every PR is normally how it's done

docs/tutorials/g3-algebra-of-space.ipynb
```diff
@@ -861,7 +861,17 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.0"
+   "version": "3.7.3"
```
eric-wieser, line 864: Can you undo the changes to this file? I don't see the point in these, I think they're just cruft from an old / customized version of jupyter

clifford/test/test_tools.py
```diff
@@ -9,15 +9,17 @@
...
+from clifford import tools
+
 
 too_slow_without_jit = pytest.mark.skipif(
     DISABLE_JIT, reason="test is too slow without JIT"
 )
 
 
-@unittest.skip("reason unknown")
+# @unittest.skip("reason unknown")
```
eric-wieser, line 20: Let's just remove this if its not needed
```suggestion
```

clifford/tools/__init__.py
```diff
@@ -130,12 +133,17 @@ def mat2Frame(A: np.ndarray,
     A : ndarray
         MxN matrix representing vectors
 
+    I: None, pseudoscalar of the frame
```
eric-wieser, line 136: The first line should be the type, not a description. Something like:
```suggestion
    I : MultiVector
```

eric-wieser:
The [docs for frame2Mat](https://clifford--389.org.readthedocs.build/en/389/api/generated/clifford.tools.frame2Mat.html#clifford.tools.frame2Mat) look wrong; perhaps some missing `:`s in the docstring?

clifford/tools/__init__.py
```diff
@@ -179,19 +187,37 @@ def mat2Frame(A: np.ndarray,
...
-    return a, layout
+    return a, I
 
 
-def frame2Mat(B, A=None, is_complex=None):
+def frame2Mat(B, A=None, I=None, is_complex=None):
+    '''
+    convert a list of vectors to a matrix
+
+    B : list
```
eric-wieser, line 197: This is missing the `Parameters` heading which sphinx needs

PR #395, opened 2021-03-03
Kill conda test with python 3.5

hugohadfield:
This is just the python3.5 conda test run, we still run 3.8 with conda and pip with 3.5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,6 @@ x-clifford-templates:
             IPython \
             h5py;
           conda install -c conda-forge sparse;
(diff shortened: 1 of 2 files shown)


eric-wieser:
For reference, this is motivated by wanting to bump to numpy 1.17, for which no conda binary exists on 3.5.
The last version with a conda binary is https://anaconda.org/anaconda/numpy/files?version=1.15.1.

eric-wieser (approved):
Let's wait to see that the CI runs at least understand their config files before merging

PR #397, opened 2021-03-04
Add rng to test_g3c_tools

hugohadfield:
Looks like I missed this whole file in the last sweep


eric-wieser (approved):
I'm going to trust CI on this one. It's only test code anyway, if we missed one CI will tell us eventually.

Issue #399, opened 2021-06-08
leftLaInv fails for complex valued multivectors


hugohadfield:
It looks like a set item type error in numba.

Minimum reproducer:
```
import clifford as cf
l,b = cf.Cl(2)
(l.randomMV()*1j).leftLaInv()
```

Stack trace:
```
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function setitem>) found for signature:
 
 >>> setitem(array(float64, 2d, C), UniTuple(int64 x 2), complex128)
 
There are 16 candidate implementations:
   - Of which 16 did not match due to:
   Overload of function 'setitem': File: <numerous>: Line N/A.
     With argument(s): '(array(float64, 2d, C), UniTuple(int64 x 2), complex128)':
    No match.

During: typing of setitem at /home/hugo/.local/lib/python3.8/site-packages/clifford/__init__.py (305)

File ".local/lib/python3.8/site-packages/clifford/__init__.py", line 305:
def _numba_val_get_left_mt_matrix(x, k_list, l_list, m_list, mult_table_vals, ndims):
    <source elided>
        i = m_list[test_ind]
        intermed[j, i] += mult_table_vals[test_ind] * x[k]
        ^

During: resolving callee type: type(CPUDispatcher(<function _numba_val_get_left_mt_matrix at 0x7f797b7bf160>))
During: typing of call at /home/hugo/.local/lib/python3.8/site-packages/clifford/_layout.py (732)

During: resolving callee type: type(CPUDispatcher(<function _numba_val_get_left_mt_matrix at 0x7f797b7bf160>))
During: typing of call at /home/hugo/.local/lib/python3.8/site-packages/clifford/_layout.py (732)


File ".local/lib/python3.8/site-packages/clifford/_layout.py", line 732:
        def leftLaInvJIT(value):
            intermed = _numba_val_get_left_mt_matrix(value, k_list, l_list, m_list, mult_table_vals, n_dims)
            ^

```

PR #400, opened 2021-06-08
Added a test for N = 7 and fixed leftLaInv complex behaviour

hugohadfield:
Fixes https://github.com/pygae/clifford/issues/399
And adds a test case to ensure we are also testing algebras >6 d
Might be more sensible to explicitly test the leftLaInv with complex numbers instead/in addition

### DIFF
--- a/clifford/__init__.py
+++ b/clifford/__init__.py
@@ -256,7 +256,7 @@ def grades_present(objin: 'MultiVector', threshold=0.0000001) -> Set[int]:
 # todo: work out how to let numba use the COO objects directly
 @_numba_utils.njit
 def _numba_val_get_left_gmt_matrix(x, k_list, l_list, m_list, mult_table_vals, ndims):
-    intermed = np.zeros((ndims, ndims))
+    intermed = np.zeros((ndims, ndims), dtype=x.dtype)
     test_ind = 0
     for k in k_list:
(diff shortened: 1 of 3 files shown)


clifford/__init__.py
```diff
@@ -256,7 +256,7 @@ def grades_present(objin: 'MultiVector', threshold=0.0000001) -> Set[int]:
 # todo: work out how to let numba use the COO objects directly
 @_numba_utils.njit
 def _numba_val_get_left_gmt_matrix(x, k_list, l_list, m_list, mult_table_vals, ndims):
-    intermed = np.zeros((ndims, ndims))
+    intermed = np.zeros((ndims, ndims), dtype=x.dtype)
```
eric-wieser, line 259: Strictly this should be `result_type(x.dtype, mult_table_vals.dtype)`, but I think numba makes that awkward and we probably don't care.
```suggestion
    # TODO: consider `dtype=result_type(x.dtype, mult_table_vals.dtype)`
    intermed = np.zeros((ndims, ndims), dtype=x.dtype)
```

eric-wieser (approved):
Looks good, let's check the new g7 test doesn't need to be marked slow

hugohadfield:
Speed-wise it looks like the g7 isn't affecting the test run times much, which is good news

PR #402, opened 2021-06-09
Switch to github actions from travis

eric-wieser:
Travis' free service supposedly ends June 15th.

I'll squash merge this if it looks good.


.github/workflows/python-package.yml
```diff
@@ -0,0 +1,169 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # fastest jobs first
+          # This job currently fails
+          # - python-version: 3.8
+          #   name: without JIT
+          #   disable_jit: 1
```
eric-wieser, line 49: It fails for the same reason it currently fails in travis

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,169 @@
...
+        python -m pip install --upgrade pip;
+        python -m pip install build
+        python -m build --sdist --wheel --outdir dist/
+    - uses: actions/upload-artifact@v2
+      with:
+        name: dist
+        path: dist
+#     - uses: pypa/gh-action-pypi-publish@v1.4.2
+#       user: arsenovic
+#       password: ieUd3c2DjrZQE+3FlqmU5FQObNWIDiL9E9G6aLs0ksEKAi5Z1t7fefXic1XHsHOZZYteycef/lZkUYg3ijwfZg2xzELeTdLef29GgUrxYuGL4MJ706UFj450Xlv9l1oH5D32OEKT2EwfxMqdrw39+N8zD5ehVyQbYM6Z3lwtCvg=
```
eric-wieser, line 168: I'm pretty sure this won't work, as the password is encrypted with a key that only travis knows, hence it being commented out.

hugohadfield (approved):

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,170 @@
...
+      run: |
+        python -m pip install --upgrade pip;
+        python -m pip install build
+        python -m build --sdist --wheel --outdir dist/
+    - uses: actions/upload-artifact@v2
+      with:
+        name: dist
+        path: dist
+    - uses: pypa/gh-action-pypi-publish@release/v1
+      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
hugohadfield, line 167:
```suggestion
      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
```

Issue #404, opened 2021-07-19
Pickling fails with NUMBA_DISABLE_JIT=1


eric-wieser:
This is why this CI run is commented out:

https://github.com/pygae/clifford/blob/659f75b407112793c6f985834d796d9f69addb78/.github/workflows/python-package.yml#L48-L51

This was brought on by a numba version increase most likely.

PR #408, opened 2021-07-20
Re-enable the no-JIT CI test

eric-wieser:
It seems the pickling hacks are broken, so I've added some more hacks on top of them to make that not matter.

What this means is that when running in no-jit mode, pickled `Layout` objects no longer contain an attempt at pickling cached operators. The pickle file continues to hold cached multiplication tables.

The regular jit mode is unaffected, as there is perhaps some value to pickling the compiled routines.

This works around #404

---

CI run: https://github.com/pygae/clifford/pull/408/checks?check_run_id=3115660735

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -45,10 +45,9 @@ jobs:
       matrix:
         include:
           # fastest jobs first
-          # This job currently fails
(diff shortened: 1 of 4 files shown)


hugohadfield (approved):
Tests pass and it looks good

PR #410, opened 2021-07-20
Run CI on Python 3.9 and state that support for Python 3.6 will not continue beyond 1.4

eric-wieser:
Note that NumPy does not build on 3.10-dev, so for now we do not test it.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -61,7 +61,7 @@ jobs:
             name: default
           - python-version: 3.9
             name: default
-          - python-version: 3.10
+          - python-version: "3.10"
             name: default
           - python-version: 3.8
             name: conda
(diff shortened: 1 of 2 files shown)


.github/workflows/python-package.yml
```diff
@@ -59,6 +59,10 @@ jobs:
             name: default
           - python-version: 3.8
             name: default
+          - python-version: 3.9
+            name: default
+          - python-version: 3.10
```
eric-wieser, line 64:
```suggestion
          - python-version: "3.10"
```

.github/workflows/python-package.yml
```diff
@@ -59,6 +59,10 @@ jobs:
             name: default
           - python-version: 3.8
             name: default
+          - python-version: 3.9
+            name: default
+          - python-version: "3.10"
```
eric-wieser, line 64:
```suggestion
          - python-version: "3.100-dev"
```

PR #412, opened 2021-08-13
numba: inverse functions, layout attributes, and better constructors

eric-wieser:
This adds numba extension support for:
* A handful of `MultiVector` methods and attributes via straightforward wrappers.
  Within these, we only do `if isinstance(self, MultiVectorType):` in the functions where we rely on the existence of `self.layout_type`.
* Some `Layout` attributes via straightforward wrappers.
* The dtype argument to `MultiVector`


hugohadfield (approved):
This looks useful and sensible, and all the tests pass

PR #414, opened 2021-09-02
numba: fix overload_call on numba >= 0.54.0

eric-wieser:
xref https://github.com/numba/numba/pull/7061.

We were burnt by relying on implementation details, but as https://github.com/numba/numba/pull/5886 never quite worked we had no choice.


eric-wieser:
CI has gone from red to green, so let's merge this.

PR #438, opened 2024-12-08
Fix GitHub workflow CI

trundev:
Currently, CI fails because of following reasons:

- Workflow job "lint" job, step "Run flake8" detects some code quality issues
- Workflow uses python 3.5, which is no longer supported by GitHub actions runner
- tools/rotor_estimation.py imports `scipy.e`, which is no longer present (see https://github.com/scipy/scipy/commit/6e207a1)

This change also avoids some warnings about outdated GitHub actions versions.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -14,12 +14,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: [3.11]
     name: "lint | Python ${{ matrix.python-version }}"
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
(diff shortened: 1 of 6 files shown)


.github/workflows/python-package.yml
```diff
@@ -14,12 +14,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: [3.11]
```
eric-wieser, line 17:
```suggestion
        python-version: ["3.11"]
```

clifford/_conformal_layout.py
```diff
@@ -66,7 +66,7 @@ def up(self, x: MultiVector) -> MultiVector:
                 new_val = np.zeros(self.gaDims)
                 new_val[:len(old_val)] = old_val
                 x = self.MultiVector(value=new_val)
-        except(AttributeError):
+        except (AttributeError):
```
eric-wieser, line 69:
```suggestion
        except AttributeError:
```
I think this was left from python 2 compatibility

clifford/cga.py
```diff
@@ -399,7 +399,7 @@ def __init__(self, cga, *args) -> None:
                     arg = float(arg)
 
             if arg < 0:
-                raise(ValueError('dilation should be positive'))
+                raise (ValueError('dilation should be positive'))
```
eric-wieser, line 402:
```suggestion
                raise ValueError('dilation should be positive')
```

.github/workflows/python-package.yml
eric-wieser: (and throughout; otherwise when we eventually reach `3.20`, yaml will parse it as `3.2`!)

trundev: That's true, 3.20 is the same as 3.2, i.e. must use "3.20".
BTW, this is valid for "3.10" - I skipped it, but it could soon become the "legacy" version.

So, quotes are mandatory for versions that ends in zeros, otherwise - recommended. I suggest quoting all the versions. Is this ok?

PR #441, opened 2024-12-12
Simplify the test for swapped operand operations

trundev:
This is a follow up for PR#437.

The pytest `TestBasicAlgebra.test_swapped_operands()` is too complicated. As suggested by @eric-wieser, the validation can be made by using some picked /convenient numbers from `g3` only.


eric-wieser (approved):
Thanks! I'll refrain from merging since you've marked this as draft.

